PR #280, opened 2015-03-30
Added some tests for PatchBuilder replaceInsert

nikos:
With the current situation of the `b3_0` sources, it looks like there are no tests for the straight replaceInsert functionality of the patch builder (EDIT: there seems to be one test under `test-complete` directory called `TestPartialUpdate` which replace/inserts on array elements). In a recent project I found out, that the behavior of the patch builder does differ if the node does exist (replace) vs. if does not yet exist (insert: will put the whole fragment instead of the selectPath). This behavior is for the application developer a bit inconvenient, since you have to check first wether the node does exist or not, and then to decide how you structure the fragment node in question.

NOTE: Therefore this test case highlights the current problem and the second assertion (L. 357, "please make me not nested") will fail.

### DIFF
--- a/src/test/java/com/marklogic/client/test/JSONDocumentTest.java
+++ b/src/test/java/com/marklogic/client/test/JSONDocumentTest.java
@@ -325,6 +325,43 @@ public class JSONDocumentTest {
 	}
 	
 	@Test
+	public void testXPathReplaceInsertPatch() throws IOException {
+		String docId = "/test/testXPathReplaceInsertPatch.json";
+		ObjectMapper mapper = new ObjectMapper();
+		ObjectNode sourceNode = makeContent(mapper);
+
+		JSONDocumentManager docMgr = Common.client.newJSONDocumentManager();
+		docMgr.write(docId, new JacksonHandle().with(sourceNode));
+
+		// ~~ Prepare the patch
+		DocumentPatchBuilder patchBldr = docMgr.newPatchBuilder();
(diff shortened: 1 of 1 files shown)


nikos:
Thanks @sammefford for bringing up the imports, I did fix those with 245a798c7d3c411714e0a149acccb7fc3bfd88eb
And yes, I did work on the b3_0 branch, for future PRs I will switch to the release-3.0.1 branch then.

sammefford:
Actually, the working development branch is 'develop'.  That's the one to use.  I can always back-port to other versions you specify where the fix is also needed.  We don't usually start work on a maintenance branch then push the changes to develop.

sammefford:
Also, as I mentioned on #280, based on the email discussion thread yesterday which I summarized, it sounds like you were able to get an answer on why the behavior is the way it is currently, and a viable workaround. Is there any need for this pull request still?

nikos:
In my opinion it would be great if the current behavior of replace-insert for patching a property on top-level in JSON documents is documented as a test case. To put it with Erik's words:

> "But it is unfortunate and confusing that replace-insert only does what you expect for XML."

If in a later release the behavior will be changed, we would then of course adapt the test case as well.

sammefford:
If I understand correctly, you're recommending we add an example to show the recommended approach for this use case.  We'll do that in the REST docs since it is the same principles for REST, Java, and Node.js.  So I'll close this issue and #280 WRT Java.  If you'd like to track progress on the docs change, please file a docs bug in bugtrack.

Issue #249, opened 2015-01-29
DocumentPatchBuilder.replaceValue cannot replace boolean values


grechaw:
This is probably actually a bug in the REST API but I found it in Java.

```
patchBuilder.replaceValue("/accepted", true);
patchBuilder.replaceValue("/accepted", new Boolean(true));
```

Both of these insert a string value into the JSON property rather than a boolean.

A workaround is to use replaceFragment
patchBuilder.replaceFragment("/accepted",  true)   

but this is a hackish workaround -- it works because the string "true", unquoted, evaluates to a boolean in constructing the patch.

grechaw:
I neglected this bug beyond code freeze, as I did not see it in my regular triage of 8.0-2 bugs.  I think I have to set it to the next version, as QA really doesn't want any new fixes for 8.0-2 anymore in the server.

grechaw:
I'll make a REST API bug for this.

grechaw:
So this is not at all a REST API bug.  It's a bug with any non-string value in replaceValue.
I've got tests written for the fix, and will have no problem delivering the fix.

grechaw:
I just assigned @sammefford  and back to me so that you'll have visbility

grechaw:
I've got a fix to review.

grechaw:
This is all merged in, should be pretty straightforward test, as the unit test shows how.
@sammefford , isn't this in 3.0.3?   I saw that you changed the milestone, just curious about that

grechaw:
Committed to develop and 3.0-develop

georgeajit:
Added tests for PatchBuilder's replaceValue on boolean for JSON and XML docs to TestPartialUpdate.java.

The tests have both boolean primitives and Boolean object.

PR #289, opened 2015-04-30
Issue #249 fixes replaceValue with numbers or booleans

grechaw:
This patch fixes issue 249 by handling replaceValue numbers, boolean, and null correctly.


grechaw:
Hi Sam, let me know whether you think this fix is ready to go or not.  I think its pretty straightforward

PR #323, opened 2015-07-15
Allow error propagation from server eval 500s

cambridgesemantics:
For example, the errorCode in variable "failure" might be "XDMP-SPQLNOSUCHGRAPH", but when extractErrorFields(response) is called a second time, the errorCode that ends up in the stack trace is null.


sammefford:
Good catch.  Thanks!

PR #333, opened 2015-07-28
Testing updates (USING) and combination query modifications

grechaw:
These commits add unit tests for USING and fix an assertion that changed (whitespace) with fast serialization of ntriples.


xquery:
this patch looks good to me

sammefford:
Charles, as we discussed I took your changes and generalized them so we can now merge a new qtext or options in addition to sparql into a combined query.  They also use the same method name "combine" as the other overloads in CombinedQueryBuilder.

grechaw:
Great!

PR #417, opened 2016-05-30
Update README.md

sashamitrovich:
Bumped up the client versions in the maven and gradle examples.


jedesah:
Ok, this PR still being open scares the hell out of me.

I think though that it is subsumed by #427 which I just opened.

sammefford:
Thank you both, sashamitrovich, and jedesah, for the contributions!  I'll close this one and we'll proceed with the more recent one.

PR #419, opened 2016-06-10
Adjusting branch names from dev to develop

robsman:
Replacing references to nonexistent branch _dev_ with _develop_.


sammefford:
Thanks Rob!

PR #422, opened 2016-07-07
Added missing extract-resolution QName

ableasdale:
Added a missing QName which was causing a Warn level log message due to its absence 

```
Unexpected metrics element {http://marklogic.com/appservices/search}extract-resolution-time
```


sammefford:
This PR was merged into develop

PR #430, opened 2016-07-15
Fix handling of lang literals (in Jena 3 they have both type and lang)

MichalWozniakSmartlogic:
This is a fix for incorrect handling of literals with language tags passed as SPARQL bindings.

Without the fix, when passing a SPARQL binding for a literal with language tag, the driver would ignore the language and do not send it to the server in the REST API call.

In Jena 2 such literals did not have any type, so that the original code worked properly (in master branch).

In Jena 3 they have both the xsd:langString type AND the language tag, so the code needs to be fixed to take that into account. This fix is only need for Jena 3 version of this library (develop branch).

### DIFF
--- a/src/main/java/com/marklogic/client/semantics/RDFTypes.java
+++ b/src/main/java/com/marklogic/client/semantics/RDFTypes.java
@@ -21,6 +21,7 @@ package com.marklogic.client.semantics;
  */
 public enum RDFTypes {
     STRING             ("string"),
+    LANGSTRING         ("langString"),
     BOOLEAN            ("boolean"),
     DECIMAL            ("decimal"),
     INTEGER            ("integer"),


grechaw:
Hi @MichalWozniakSmartlogic , I'll investigate this PR.    From looking at the REST docs, this looks like a straight-ahead bug in the Java Client API.  Was that your assessment too?

grechaw:
We will want to add a unit test to verify.  I'll take a look at the marklogic-jena PR

pit3k:
I've reviewed status of it again, and in fact for our use-case, no changes to this project are needed, as long as <https://github.com/marklogic/marklogic-jena/issues/42> is merged in marklogic-jena project (which is a serious defect).

Regarding changes in this pull-request, the change in JerseyServices is not needed at all, but the change in RDFTypes might be important for some usages. So, I've adjusted the pull request to include only this change.

pit3k:
For completeness, here is the stacktrace from the error, we're getting without neither of the pull-requests applied, when invoking MarkLogicDatasetGraph.bindObject with a node that is a literal with a language tag:

```
java.lang.IllegalArgumentException: No enum constant com.marklogic.client.semantics.RDFTypes.LANGSTRING
java.lang.Enum.valueOf(Enum.java:238)
com.marklogic.client.semantics.RDFTypes.valueOf(RDFTypes.java:22)
com.marklogic.semantics.jena.MarkLogicDatasetGraph.bindObject(MarkLogicDatasetGraph.java:152)
com.smartlogic.topbraid.marklogic.update.MarklogicUpdateClient$1.processSparql(MarklogicUpdateClient.java:105)
```

grechaw:
@sammefford , as this issue was fixed in the marklogic-jena codebase, you can close this PR.

PR #432, opened 2016-07-27
Providing a portable variant to initialize (and teardown) a test database

robsman:
Having to use shell scripts may confuse non-Linux/Mac contributors.

Additionally, this will use the classpath as modeled in pom.xml and avoids the
need to keep shell scripts in sync. If this gets accepted, may the
`bootstrap-test.sh`and `bootstrap-test-cygwin.sh` scripts become obsolete?


sammefford:
Looks good to me!

PR #443, opened 2016-08-16
Use equals() rather than comparing String object using == or !=

hansenmc:
- Changed == and != to use equals() when comparing strings
- Added a unit test with positive/negative tests
- Added Mockito to test dependencies


sammefford:
this PR was merged into develop in 11e89ef

PR #444, opened 2016-08-16
May fail to close stream

hansenmc:
- May fail to close stream, especially if exceptions occur.
- Changed to use try-with-resources for more concise code that will
ensure that streams are closed.


sammefford:
this PR was merged into develop in d5bccba

PR #447, opened 2016-08-17
Code to interface and use diamond operator to eliminate redundant type args

hansenmc:
- Declare objects using the interface, rather than an implementation.
- Eliminate redundant type arguments. Use the diamond operator for the
right hand side of the object assignment.


sammefford:
I'm thinking we don't want to change the signature of public methods as clients may be using those and we don't want to make backward-incompatible changes lightly.  It's true there was no reason for getValuesMap to return a HashMap--those should have returned a Map, but that's water under the bridge now.  So I'll accept as many of these changes as I can, but those I don't think we should make.  Also, for some reason one of the changes in JerseyServices didn't compile, so I had to revert that.

sammefford:
this PR was merged into develop in f4abcd9 and a880d80

PR #495, opened 2016-10-10
fix #125 - breakout WriteHostBatcher.flush to flushAsync and flushAndWait

### LINKED ISSUE #125: Missing the implementation for Logging?
I see all the Managers have a way to startLogging() and It would be nice/needed for PojoRepository to have startLogging() and stopLogging() methods implemented.

sammefford:
This fixes https://github.com/marklogic/data-movement/issues/125

Some early feedback with WriteHostBatcher was that sometimes one might want to flush without blocking for everything to complete.  Some were unsure whether to expect flush to block or not.  We decided to  remove flush and replace it with flushAndWait and flushAsync.  The method names hopefully illustrate what they do.  We discussed this in feature team meeting, but it's been a *long* time back.  Let me know if this raises any concerns.

Charles, can you merge this into entity services, then change any of your calls to writeHostBatcher.flush() to be flushAndWait() instead?  

Ajit, can you do the same and merge in this pull request when you've updated the functional tests to not use flush() any longer?

### DIFF
--- a/src/main/java/com/marklogic/client/datamovement/WriteHostBatcher.java
+++ b/src/main/java/com/marklogic/client/datamovement/WriteHostBatcher.java
@@ -32,8 +32,8 @@ import com.marklogic.client.io.marker.DocumentMetadataWriteHandle;
  * time enough documents are added to make a batch, the batch is added to an
  * internal queue where the first available internal thread will pick it up and
  * write it to the server.  Since batches are not written until they are full,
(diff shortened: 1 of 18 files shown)


sammefford:
Charles & Ajit, could you each weigh in on whether you're ready to accept this pull request?  Ajit, if you're ready, could you merge it in?

grechaw:
Go for it.    Merging may break entity services regression, but I'll adjust my code today.

Issue #592, opened 2016-12-08
Cannot parse JSON Error when processing a long response


laquereric:
Cannot parse JSON Error when processing a long response

```
com.marklogic.client.MarkLogicIOException: Cannot parse JSON '[{"ICS_codeData":{"midbData":{"co' for fn:doc("/content/GUIDE://1972/4442a1b1-2107-4796-b2da-7bf722d05fc1/ndl-w/u/|||/a2a6f9a4-3261-421b-a309-00631c482ebc.json")
        at com.marklogic.client.io.SearchHandle$MatchDocumentSummaryImpl.populateExtractedResult(SearchHandle.java:607) ~[java-client-api-3.0.3.jar:na]
        at com.marklogic.client.io.SearchHandle$MatchDocumentSummaryImpl.getExtracted(SearchHandle.java:567) ~[java-client-api-3.0.3.jar:na]
        at MarkLogicDatabaseClientTestCase.queryForCodePage(MarkLogicDatabaseClientTestCase.java:127) [TestCase/:na]
        at MarkLogicDatabaseClientTestCase.main(MarkLogicDatabaseClientTestCase.java:162) [TestCase/:na]
Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected end-of-input: was expecting closing '"' for name
 at [Source: [{"ICS_codeData":{"midbData":{"co; line: 1, column: 67]
        at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1581) ~[jackson-core-2.6.5.jar:2.6.5]
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:533) ~[jackson-core-2.6.5.jar:2.6.5]
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportInvalidEOF(ParserMinimalBase.java:470) ~[jackson-core-2.6.5.jar:2.6.5]
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._parseName2(ReaderBasedJsonParser.java:1447) ~[jackson-core-2.6.5.jar:2.6.5]
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._parseName(ReaderBasedJsonParser.java:1431) ~[jackson-core-2.6.5.jar:2.6.5]
[...]

PR #594, opened 2016-12-08
Fixes cannot parse JSON error from Support Case 17586

laquereric:
Fix for issue #592 on branch develop


sammefford:
Fixing #592 and #598 instead with this: https://github.com/marklogic/java-client-api/commit/1bfe12a7527737bebb4abc01d3ded0ff150a67b5

PR #716, opened 2017-03-31
ensure that FileInputStream is closed to avoid resource leakage

hansenmc:
Changed test for `instanceof File` to `instanceof FileInputStream`, so that
FileInputStreams will be closed.

Currently, this condition always fails and FileInputStreams are not
being closed when they should.


sammefford:
@hansenmc I can't merge this without a unit test.  I'll keep it so I can add a unit test when I get time.  If you beat me to it, I'll be able to merge this.

hansenmc:
Actually, now that I take a closer look - the `close()` is unnecessary and that code block can be removed altogether. So, rather than a possible memory leak, it's unnecessary, closing the already closed `FileInputStream`.

If a `File` is passed in, then [content is assigned as a `new FileInputStream()`](https://github.com/marklogic/java-client-api/blob/develop/src/main/java/com/marklogic/client/impl/HandleAccessor.java#L114), which then uses [try-with-resources to read with an `InputStreamReader`](https://github.com/marklogic/java-client-api/blob/develop/src/main/java/com/marklogic/client/impl/HandleAccessor.java#L118), which will automatically invoke `close()`.

So, whether a File or FileInputStream were sent in, they should always be read with the InputStreamReader and automatically closed. 

I added a specific test to demonstrate that `FileInputStream` is indeed being closed automatically and removed that code block. I couldn't think of a way to demonstrate/test File, but if you look at that section of code, it should be apparent.

sammefford:
Analysis looks perfect to me.  merging.  Thanks!

PR #717, opened 2017-03-31
prevent infinite recursion and stack overflow

hansenmc:
The `writeRuleAs(String ruleName, Object ruleSource)` method was
invoking itself, which would recurse infinitely, until a stack overflow
is thrown.

It appears that it should be invoking `writeRule()` instead.


sammefford:
Thanks!  This commit looks right.  But I won't be able to integrate it until I get time to add a unit test.  If you have time to add a unit test for it, I could integrate right away.

hansenmc:
I had started writing a unit test, but using a Mockito mock service was more involved than I had expected. I probably just need my own RESTService impl that supports putValue() and exists(). I'll take a look.

sammefford:
Let me caveat: when I say "unit test" in this project, I technically mean integration test.  We like to run our tests against the server to ensure that things are working as expected end-to-end.  And it helps us catch things when changes in the server break our tests.  I think the only place we mock anything in our "unit tests" is something you contributed.  Would you be comfortable adding a test to [AlertingTest.java](https://github.com/marklogic/java-client-api/blob/develop/src/test/java/com/marklogic/client/test/AlertingTest.java)?  Or even better, just take one of the existing tests in there that calls writeRule and change it to use writeRuleAs.

hansenmc:
I'd be happy to add a test to AlertingTest, but those integration tests are currently failing when I run them. 

There seems to be a permissions issue. Any idea what roles the user that the tests use are supposed to have? Maybe the test bootstrap is missing something?

I'm running MarkLogic 9.0-20170113 and have a clean checkout of the develop branch.

I have uninstalled the test database and appserver using `bootstrapper.teardown()` and then followed the instructions to run:
 - mvn test-compile
- mvn exec:java@test-server-init
- mvn test

However, results are:
>Tests run: 609, Failures: 0, Errors: 540, Skipped: 0

I see various failures in the run:
> com.marklogic.client.test.SuggestTest  Time elapsed: 0.012 sec  <<< ERROR!
com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to delete documents. Server Message: You do not have permission to this method and URL.
	at com.marklogic.client.impl.JerseyServices.deleteDocument(JerseyServices.java:603)

> testSemUuid0Exec(com.marklogic.client.test.PlanGeneratedTest): Local message: User is not allowed to apply resource at rows. Server Message: You do not have permission to this method and URL.

> testAggregates(com.marklogic.client.test.TuplesHandleTest)  Time elapsed: 0.007 sec  <<< ERROR!
com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to write /config/query. Server Message: You do not have permission to this method and URL.
	at com.marklogic.client.impl.JerseyServices.putPostValueImpl(JerseyServices.java:2817)

>testFnHead1Exec(com.marklogic.client.test.PlanGeneratedTest)  Time elapsed: 0.011 sec  <<< ERROR!
com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to apply resource at rows. Server Message: You do not have permission to this method and URL.
	at com.marklogic.client.impl.JerseyServices.checkStatus(JerseyServices.java:4305)

hansenmc:
I removed the users and re-ran the bootstrap.xqy methods to create users and roles. I encountered errors for two roles do not exist, temporal-document-protect and temporal-document-wipe. I'm assuming they are bleeding edge only available in the latest MarkLogic builds?

After commented out these lines in the bootstrap.xqy:
>    local:security-eval(
        'sec:privilege-add-roles("http://marklogic.com/xdmp/privileges/temporal-document-protect", "execute", "rest-admin")'),
    local:security-eval(
        'sec:privilege-add-roles("http://marklogic.com/xdmp/privileges/temporal-document-wipe", "execute", "rest-admin")'),

However, I still saw permissions errors, such as this:
>  testTransactions(com.marklogic.client.test.SPARQLManagerTest): Local message: User is not allowed to apply resource at /graphs/sparql. Server Message: You do not have permission to this method and URL.

amd now I'm seeing REST-INVALIDPARAM errors for the return type `object`

>  testFnTrueExpr0Exec(com.marklogic.client.test.PlanGeneratedTest): Local message: failed to apply resource at rows: Bad Request. Server Message: REST-INVALIDPARAM: fn.error(null, 'REST-INVALIDPARAM', 'unknown output type: '+outputType); -- Invalid parameter: unknown output type: object

But, the AlertTest tests execute, so I can add and verify.

hansenmc:
One thing that I noticed, wasn't sure if this is another bug or intended behavior...

If `writeRuleAs(ruleName, ruleSource)` invokes `[writeRule(ruleName, writeHandle)`](https://github.com/marklogic/java-client-api/blob/develop/src/main/java/com/marklogic/client/impl/RuleManagerImpl.java#L141) and the writeHandle is a RuleDefinition with a name, then it will be saved with the RuleDefinition name, not the name provided when writeRuleAs was invoked.

I would think that the purpose of `writeRuleAs()` is to ensure that it is saved with the name provided and that we would want to clear the `RuleDefinition.name` prior to invoking `writeRule()`. I'm updating writeRuleAs() with the following, to ensure that it is saved with the name specified:

    if (sourceHandle instanceof RuleDefinition) {
      ((RuleDefinition) sourceHandle).setName(ruleName);
    }

PR #840, opened 2017-10-06
Adding placeholders for snapshot and release repositories

itsshivaverma:
Hi @grechaw and @sammefford

This PR will update the pom so that we can specify the snapshot and release nexus server URLS either at the command-prompt or through settings.xml

Thanks
Shiva


grechaw:
Thanks Sam!

PR #928, opened 2018-05-18
Added ProgressListener

rjrudin:

This is a reusable listener that I created on another project using DMSDK where I wanted to keep track of progress while running a large query.

Yes, the develop branch.

Yes, and I added ProgressListenerTest. 

4.0.4

9.0-5

1.8.0_131

* OS and version

N/A

N/A

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2015-2018 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.client.datamovement;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+/**
+ * <p>Reports on progress as batches are processed by sending an instance of the nested ProgressUpdate interface to
+ * instances of java.util.function.Consumer.</p>
+ *
+ * As an example, one approach would be to log the progress as a string:
+ *
+ * <pre>{@code
+ *      QueryBatcher queryBatcher = moveMgr.newQueryBatcher(query)
+ *         .withConsistentSnapshot()
+ *         .onUrisReady(
+ *           new ProgressListener()
+ *             .onProgressUpdate(progressUpdate -> {
+ *               System.out.println(progressUpdate.getProgressAsString());
+ *             })
+ *         );
+ *
+ *     JobTicket ticket = moveMgr.startJob(queryBatcher);
+ *     queryBatcher.awaitCompletion();
+ *     moveMgr.stopJob(ticket);
+ * }</pre>
+ *
+ * The method {@link #withTotalResults(long) withTotalResults} can be used to inform the listener of the total number
+ * of expected results; this would likely have been determined by first using a QueryManager with a page length of zero
+ * to determine the number of results for a query. Given the total results, each ProgressUpdate object will be aware of
+ * how close to complete the QueryBatcher is.
+ */
+public class ProgressListener implements QueryBatchListener {
+
+	private static Logger logger = LoggerFactory.getLogger(ProgressListener.class);
+
+	private List<Consumer<ProgressUpdate>> consumers = new ArrayList<>();
+	private AtomicLong resultsSoFar = new AtomicLong(0);
+	private long startTime;
+	private long totalResults;
+
+	public ProgressListener() {
+	}
+
+	/**
+	 * Use this constructor for when the total number of results isn't known ahead of time.
+	 *
+	 * @param consumers
+	 */
+	public ProgressListener(Consumer<ProgressUpdate>... consumers) {
+		this(0, consumers);
+	}
+
+	/**
+	 * Use this constructor for when the total number of results is known ahead of time. The ProgressUpdate objects that
+	 * are emitted will have more interesting information as a result.
+	 *
+	 * @param consumers
+	 * @param totalResults
+	 */
+	public ProgressListener(long totalResults, Consumer<ProgressUpdate>... consumers) {
+		this.totalResults = totalResults;
+		for (Consumer<ProgressUpdate> consumer : consumers) {
+			this.consumers.add(consumer);
+		}
+	}
+
+	public ProgressListener withTotalResults(long totalResults) {
+		this.totalResults = totalResults;
+		return this;
+	}
+
+	public ProgressListener onProgressUpdate(Consumer<ProgressUpdate> consumer) {
+		this.consumers.add(consumer);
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
```diff
@@ -0,0 +1,189 @@
...
+	/**
+	 * Protected so that a subclass can override how a consumer is invoked, particularly how an exception is handled.
+	 *
+	 * @param consumer
+	 * @param progressUpdate
+	 */
+	protected void invokeConsumer(Consumer<ProgressUpdate> consumer, ProgressUpdate progressUpdate) {
+		try {
+			consumer.accept(progressUpdate);
+		} catch (Throwable t) {
```
grechaw: Normally I'd want to distinguish more between OK errors and ones that need to be rethrown.  It might help somebody testing and debugging later on  if Throwable is refined a bit.  Up to you.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
```diff
@@ -0,0 +1,189 @@
...
+					return text + "; " + rate + " records/s";
+				}
+				return text;
+			}
+
+			return String.format("Progress: %d results so far; time %fs", queryBatch.getJobResultsSoFar(), timeSoFarInSeconds);
+		}
+
+		@Override
+		public boolean isComplete() {
```
grechaw: what if the query result is 0?

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/ProgressListenerTest.java
```diff
@@ -0,0 +1,175 @@
...
+		listener.initializeListener(null);
+		// Ensure that the total amount of time will be greater than zero for when it's tested below
+		Thread.sleep(100);
+
+		listener.processEvent(new FakeQueryBatch(2));
+		assertFalse(consumer.lastProgressUpdate.isComplete());
+
+		listener.processEvent(new FakeQueryBatch(4));
+		assertTrue(consumer.lastProgressUpdate.isComplete());
+
```
grechaw: another test of FakeQueryBatch(0) might expose the problem with above isComplete()

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
```diff
@@ -0,0 +1,189 @@
...
+	private AtomicLong resultsSoFar = new AtomicLong(0);
+	private long startTime;
+	private long totalResults;
+
+	/**
+	 * Use this constructor for when the total number of results isn't known ahead of time.
+	 *
+	 * @param consumers
+	 */
+	public ProgressListener(Consumer<ProgressUpdate>... consumers) {
```
vivekmuniyandi: Can we have the consumers registered via a new onProgressUpdate() method rather than taking the consumers in the constructor. That provides a little more flexibility when it comes to adding Consumers and it would be consistent with the other listeners. Something along the lines of 
```
public ProgressListener onProgressUpdate(Consumer<ProgressUpdate> consumer) 
```
and return ProgressListener so that we can chain them. Similarly for totalResults as well,
```
public ProgressListener withTotalResults(long totalResults) 
```

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
rjrudin: I followed what ExportListener is doing as it iterates over consumers - it catches a Throwable and logs it as well.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
rjrudin: Good point, I'll model it that way - I'll keep the constructor arg approach as well.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
rjrudin: Do you mean totalResults? If that's zero, this will always return false since it doesn't know when the job completes. 

This is based on some work on a project where I first use QueryManager and a page length of zero to get a count of matching results. Then I kick off the QueryBatcher, feeding totalResults into the progress listener so it knows when "done" is arrived at.

grechaw: ah, with that design in mind this makes more sense.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
rjrudin: Added chaining methods

rjrudin:
Thanks @jmakeig - while we're on the subject, Warriors in how many?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
vivekmuniyandi: Thanks!

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
```diff
@@ -0,0 +1,202 @@
...
+			logger.error("Exception thrown by a Consumer<ProgressUpdate> consumer: " + consumer + "; progressUpdate: " + progressUpdate, t);
+		}
+	}
+
+	/**
+	 * Captures data of interest for a progress update.
+	 */
+	public interface ProgressUpdate {
+
+		String getProgressAsString();
```
vivekmuniyandi: @rjrudin A belated thought! Do we always expect a String here? Can we use generics here and return a generic object so that just like how SimpleProgressUpdate can return a String and other implementations of ProgressUpdates can return whatever they want! What do you think?

Apologies for having this comment so late.

kcoleman-marklogic:
If we think this is a generally useful listener, please beef up the class javadoc to be in line with things like DeleteListener and ExportListener. For example, include an example of how to use it, and anything else the user needs to know to make best use of it.

vivekmuniyandi:
Also @rjrudin can you make the PR against post4.1 branch? I have created a new branch for tasks/bugs  that should go in post 4.1 as we have 4.1 planned entirely for API-First. The branch name is "post4.1"

Sorry for the runaround.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
```diff
@@ -0,0 +1,202 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+/**
```
kcoleman-marklogic: The class level javadoc needs more substance. At least an example of how to use it. If there are things the user should know, include that, too. For an example of what I'm talking about, take a look at DeleteListener or ExportListener.

rjrudin: Got it, I'm working on ExportListener-like class comments now.

rjrudin:
@vivekmuniyandi I'll add the class comments and close this and submit a new one against the branch you mentioned.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ProgressListener.java
rjrudin: I'm holding off on this for now, only because I don't think there's a strong use case for ProgressUpdate as an interface itself. Using a different implementation of it requires subclassing ProgressListener and overriding a protected method. If a client does that, then can always write a Consumer<ProgressUpdate> that casts the ProgressUpdate to their own concrete implementation in order to access any other functionality.

vivekmuniyandi: Makes sense! Thanks Rob!

Issue #943, opened 2018-06-19
integration of base classes for API First into internal libraries


ehennum:
Subtask of internal issue DHFPROD-619

ehennum:
Sending over to you, Vivek, for the OKHttpServices integration.  When that's done, please send back to me for cleanup on the base class for API First.

vivekmuniyandi:
Done with the integration and sending this back to you Erik!

ehennum:
The Java test classes generated from the *.api files by API First will need to be regenerated.  

After that that, if there are no regressions, this issue can be closed.

PR #953, opened 2018-07-03
#943 - API First Base class integration to OkHttpServices

vivekmuniyandi:
This pull request incorporates API First Base class integration into OkHttpservices. 

All the unit tests in both java client api and development tools subprojects were successful.


vivekmuniyandi:
I will merge it for now so that you can work on it. I will make those changes and do my testing and then commit the changes later. Thanks Erik for helping me through this.

Issue #965, opened 2018-07-23
Enable QueryBatcher to take in a RawCtsQueryDefinition 


vivekmuniyandi:
Right now QueryBatcher can't take in a Raw Cts Query and process results based on the Cts query passed. We need to enable QueryBatcher to take in a RawCtsQueryDefinition in order to accomplish this.

vivekmuniyandi:
Trunk - Committed revision 291425.
b9 - Committed revision 291424.

vivekmuniyandi:
This has a server dependency. Hence 4.1.1 should be aligned with 9.0.7 or it should be released later.

georgeajit:
Running a test with ```RawCtsQueryDefinition```  on a QuerBatcher instance produces ```REST-UNSUPPORTEDPARAM```

The stack trace is:

```
com.marklogic.client.datamovement.QueryBatchException: com.marklogic.client.FailedRequestException: Local message: failed to apply resource at internal/uris: Bad Request. Server Message: REST-UNSUPPORTEDPARAM: (err:FOER0000) Endpoint does not support query parameter: invalid parameters: <cts:word-query xmlns:cts
	at com.marklogic.client.datamovement.impl.QueryBatcherImpl$QueryTask.run(QueryBatcherImpl.java:656)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
Caused by: com.marklogic.client.FailedRequestException: Local message: failed to apply resource at internal/uris: Bad Request. Server Message: REST-UNSUPPORTEDPARAM: (err:FOER0000) Endpoint does not support query parameter: invalid parameters: <cts:word-query xmlns:cts
	at com.marklogic.client.impl.OkHttpServices.checkStatus(OkHttpServices.java:4280)
	at com.marklogic.client.impl.OkHttpServices.postResource(OkHttpServices.java:3264)
	at com.marklogic.client.impl.OkHttpServices.postResource(OkHttpServices.java:3210)
	at com.marklogic.client.impl.OkHttpServices.postResource(OkHttpServices.java:3201)
	at com.marklogic.client.impl.OkHttpServices.uris(OkHttpServices.java:2954)
	at com.marklogic.client.impl.QueryManagerImpl.uris(QueryManagerImpl.java:179)
[...]

PR #978, opened 2018-08-01
#965 - Support Raw Cts Query Definition in QueryBatcher

vivekmuniyandi:


kcoleman-marklogic:
Vivek, is this going into tomorrow's release??

vivekmuniyandi:
Kim, no this is for 4.1.1 - the next release. We are branching the release out of 4.1 branch and develop has changes which do not go into 4.1.

vivekmuniyandi:
Thanks for catching that out. I was actually sending it directly. I needed the serialize method only for logging. Adding a method just for logging seems overkill. I will remove them and call the implementation's method for the log. Thanks Erik.

PR #979, opened 2018-08-02
Fix XML serialization in DOMWriter

ryanjdew:
Complex XML with namespaces were getting serialized incorrectly in the properties. This was found out by trying to write back properties that had extracted text. The root cause seemed to be parameters passed in the wrong order to a few functions. 
I believe so. This needs to be fixed in both the 3.x and 4.x releases.
Yes and I've added a unit test to cover this case. I did see some failures, but it appeared to be the same number of failures as before my changes.
4.0.4 and 3.0.8
MarkLogic 9.0-6 and 8.0-6
1.8.0_77
* OS and version
MacOS 10.13.6 
XML serialization for the metadata would create invalid XML like below:
    ```xml
    <rapi:metadata xmlns:rapi="http://marklogic.com/rest-api" xmlns:prop="http://marklogic.com/xdmp/property" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema">
      <rapi:permissions>
        <rapi:permission>
          <rapi:role-name>rest-writer</rapi:role-name>
          <rapi:capability>update</rapi:capability>
        </rapi:permission>
        <rapi:permission>
          <rapi:role-name>rest-reader</rapi:role-name>
          <rapi:capability>read</rapi:capability>
        </rapi:permission>
      </rapi:permissions>
      <prop:properties>
        <filter:text xmlns:filter="http://marklogic.com/filter">
          <zdef-1006346975:http://www.w3.org/1999/xhtml xmlns:zdef-1593109095="xmlns" xmlns:zdef-1006346975="body" zdef-1593109095:http://www.w3.org/2000/xmlns/="http://www.w3.org/1999/xhtml">
            <zdef-1303289172:http://www.w3.org/1999/xhtml xmlns:zdef-1303289172="p">
              This is a test!</zdef-1303289172:http://www.w3.org/1999/xhtml>
          </zdef-1006346975:http://www.w3.org/1999/xhtml>
        </filter:text>
        <content-type xmlns="http://marklogic.com/filter" xsi:type="xs:string">text/plain</content-type>
        <testId xsi:type="xs:string">123456</testId>
        <size xmlns="http://marklogic.com/filter" xsi:type="xs:string">15</size>
        <filter-capabilities xmlns="http://marklogic.com/filter" xsi:type="xs:string">text</filter-capabilities>
      </prop:properties>
      <rapi:quality>0</rapi:quality>
    </rapi:metadata>
    ```

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/DOMWriter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/DOMWriter.java
@@ -18,9 +18,6 @@ package com.marklogic.client.impl;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import org.w3c.dom.Attr;
 import org.w3c.dom.CDATASection;
 import org.w3c.dom.Comment;
@@ -35,7 +32,6 @@ import org.w3c.dom.Text;
 import com.marklogic.client.MarkLogicInternalException;
 
 public class DOMWriter {
-  final static private Logger logger = LoggerFactory.getLogger(DOMWriter.class);
   private XMLStreamWriter serializer;
 
   public DOMWriter() {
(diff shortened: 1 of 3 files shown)


ehennum:
@anu3990 , could you review this pull request including the unit tests? Thanks.

anu3990:
"logger" variable is never used in DOMWriter so we can remove lines - 21,22,23 and 38. Same with DocumentMetadataHandleTest, lines 35,36 and 59.

ryanjdew:
logger and corresponding imports have been removed

anu3990:
Thanks Ryan.

PR #1014, opened 2018-10-05
Adding support for constructing a DatabaseClient via configuration

rjrudin:

This copies classes from the ml-javaclient-util into java-client-api so that they can be more easily kept in sync with `DatabaseClientFactory`. These classes are used in several places (ml-gradle, DHF, NiFi) for constructing a `DatabaseClient` by calling setters on `DatabaseClientConfig`.

* Are you modifying the correct branch? 

The only branch I can really push this to right now is `4.0-master`, because the develop branch does not yet have the `ConnectionType` addition in `DatabaseClient`. Once `develop` is updated with that addition, this PR should point to `develop`. 

* Have you run unit tests?

Where'd CONTRIBUTING.md go?

4.1.1 was just released, so I think this would go into 4.2?

N/A

N/A

* OS and version

N/A

See above.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/config/ConfiguredDatabaseClientFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/config/ConfiguredDatabaseClientFactory.java
@@ -1,0 +1,14 @@
+package com.marklogic.client.config;
+
+import com.marklogic.client.DatabaseClient;
+
+/**
+ * Hides how a DatabaseClient is constructed based on the inputs in a DatabaseClientConfig object. The intent is that a
+ * client can populate any set of properties on the DatabaseClientConfig, and an implementation of this interface will
+ * determine how to construct a new DatabaseClient based on those properties.
+ */
+public interface ConfiguredDatabaseClientFactory {
+
+	DatabaseClient newDatabaseClient(DatabaseClientConfig databaseClientConfig);
+
+}
--- a/marklogic-client-api/src/main/java/com/marklogic/client/config/DatabaseClientConfig.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/config/DatabaseClientConfig.java
@@ -1,0 +1,146 @@
+package com.marklogic.client.config;
+
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.DatabaseClientFactory.SSLHostnameVerifier;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.X509TrustManager;
+
+/**
+ * Captures all the possible inputs used to construct an instance of DatabaseClient.
+ */
+public class DatabaseClientConfig {
+
+	private SecurityContextType securityContextType = SecurityContextType.DIGEST;
+	private String host;
+	private int port;
+	private String username;
+	private String password;
+	private String database;
+	private SSLContext sslContext;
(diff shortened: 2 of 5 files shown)


ehennum:
Thanks, @rjrudin , for bringing this forward.  Could the existing com.marklogic.client.DatabaseClientFactory.Bean class meet any of the requirements motivating DatabaseClientConfig?  The current DatabaseClientFactory.Bean class:

https://github.com/marklogic/java-client-api/blob/4.0-master/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java#L1011

rjrudin:
Yes, though the Bean class is missing some inputs. Perhaps it should contain all of the logic for constructing a DatabaseClient as well?

ehennum:
@rjrudin thanks for calling attention to the missing fields.  I'll create a bug for that.

I'm wondering whether we should provide the following factory:

    public static DatabaseClient DatabaseClientFactory.newClient(DatabaseClientFactory.Bean) {...}

My reasoning is that, in principle, the factory can have other configuration (such as the handle registry) that affects client creation.

Either way, we should have a bug to provide a factory method that uses the Bean for client creation.

(Bean.  Such an archaic term.)

ehennum:
I've added #1018 for comparing DatabaseClientFactory.Bean with com.marklogic.client.ext.DatabaseClientConfig

I have to retract my speculation in the previous comment. DatabaseClientFactory.Bean has the entire factory state, not just the database client configuration.  So, if it were passed to a factory method, that would introduce the potential for conflict between the state of two factories.

Instead, DatabaseClientFactory.Bean provides a zero-argument client factory method:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java#L1478

Given the @rjrudin , if we add setters for missing configuration to DatabaseClientFactory.Bean, would that meet the requirement that motivated this pull request?

rjrudin:
@ehennum I like the idea of Bean capturing all the inputs - what class would then have the knowledge of how to construct a DatabaseClient based on any set of defined inputs within Bean? That's what the ConfiguredDatabaseClientFactory implementation knows how to do.

ehennum:
@rjrudin , when @anu3990 investigated, it turned out that DatabaseClientFactory.Bean already had interfaces for all of the configuration of a DatabaseClient. Some of the authentication setters, however, were noops.  The fix will appear in the next release.

On the question, I don't understand why ConfiguredDatabaseClientFactory is needed if DatabaseClientFactory.Bean is available. DatabaseClientFactory.Bean is the class that has knowledge of how to construct a DatabaseClient when DatabaseClientFactory.Bean.newClient() is called:

http://docs.marklogic.com/javadoc/client/com/marklogic/client/DatabaseClientFactory.Bean.html#newClient--

What am I missing?

rjrudin:
@ehennum I didn't know that Bean.newClient() exists. I assume that doesn't work today because of the noop setters, but will work in the next release?

If so, I'll gladly shift to using Bean.newClient().

ehennum:
@rjrudin , Bean.newClient() worked for the legacy authentication properties but not for the refactored SecurityContext object.  @anu3990 fixed that under:

https://github.com/marklogic/java-client-api/issues/1018

rjrudin:
Thanks @ehennum  I'll close this, and once the Java client release has the above PR in it, I'll rework ml-javaclient-util so that it uses Bean under the hood.

PR #1026, opened 2018-11-08
Write before job document.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/DHFPROD-1468
* OS and version

### DIFF
--- a/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
+++ b/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
@@ -14,9 +14,7 @@ import com.marklogic.client.io.JacksonHandle;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.LocalDateTime;
 import java.util.Iterator;
 import java.util.UUID;
 
@@ -55,23 +53,17 @@ public class JobRunner {
           ObjectNode recordDef     = (ObjectNode) jobControlObj.get("record");
           
           if(jobDef.isMissingNode()) {
-         	 throw new MarkLogicIOException("job Node is missing.");
+         	 throw new MarkLogicIOException("job Node cannot be empty.");
           }
           String id = jobDef.path("id").asText();
           if(id==null || id.length()==0) {
          	 throw new MarkLogicIOException("job id cannot be empty or Null");
           }
   		JsonNode metadataNode = jobDef.path("metadata");
-		if (metadataNode.isMissingNode()) {
-			throw new MarkLogicIOException("metadata is  missing.");
-		}
-		String reconciliation = metadataNode.path("reconciliation").asText();
+		String jobMetadataValue = metadataNode.asText();
 		DocumentMetadataHandle documentMetadata = new DocumentMetadataHandle();
 		documentMetadata.withCollections("/jobs/"+id);
          
-		DateFormat sdf = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
-		Date date = new Date();
-		String dateTime = sdf.format(date);
 // TODO: permissions
 
          WriteBatcher batcher = moveMgr.newWriteBatcher()
@@ -97,9 +89,7 @@ public class JobRunner {
 
 
          ObjectLoader loader = new ObjectLoader(batcher, recordDef, getJobDirectory(), documentMetadata);
-         CSVConverter converter = new CSVConverter();
-
-
+         CSVConverter converter = new CSVConverter();  
 
          Iterator<ObjectNode> itr = converter.convertObject(csvRecords);
          if(!itr.hasNext()) {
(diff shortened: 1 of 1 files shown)


ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -37,10 +45,33 @@ public String getJobDirectory() {
...
+          }
+          String id = jobDef.path("id").asText();
+          if(id==null || id.length()==0) {
+         	 throw new MarkLogicIOException("job id cannot be empty or Null");
+          }
+  		JsonNode metadataNode = jobDef.path("metadata");
+		if (metadataNode.isMissingNode()) {
+			throw new MarkLogicIOException("metadata is  missing.");
+		}
+		String reconciliation = metadataNode.path("reconciliation").asText();
```
ehennum: Let's use the job metadata as is without looking inside it.  

Also, if the job metadata is missing, let's omit it instead of throwing an error.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -37,10 +45,33 @@ public String getJobDirectory() {
...
+		DocumentMetadataHandle documentMetadata = new DocumentMetadataHandle();
+		documentMetadata.withCollections("/jobs/"+id);
+         
+		DateFormat sdf = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+		Date date = new Date();
+		String dateTime = sdf.format(date);
```
ehennum: Or possibly LocalDateTime.now().toString()?  The Java API requires Java 8 as a minimum, so the Time library is available:
https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
 
+
+         Iterator<ObjectNode> itr = converter.convertObject(csvRecords);
+         if(!itr.hasNext()) {
+        	 throw new MarkLogicIOException("No header found.");
+         }
+         ObjectNode csvNode = itr.next();
+         
+         Iterator<String> csvHeader = csvNode.fieldNames();
+         String headerValue = csvNode.get(csvHeader.next()).toString();
```
ehennum: Isn't csvNode the first data row rather than the header row?  If so, we want to add the field names and not the field values.  To put it another way, doesn't line 111 return the value of the first field in the first data row?

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
+         
+         ObjectNode beforeDocRoot = objectMapper.createObjectNode();
+         beforeDocRoot.put("job id", id);
+         beforeDocRoot.put("reconciliation",  reconciliation);
+         beforeDocRoot.put("Date",  dateTime.toString());
+         beforeDocRoot.put("headerValue",  headerValue);
+         
+         JacksonHandle jacksonHandle = new JacksonHandle(beforeDocRoot);
+         
+         jsonMgr.write(beforeDocId, beforeDocumentMetadata, jacksonHandle);
```
ehennum: The before job document should be written before the call to moveMgr.startJob(), which will start writing JSON documents for the job.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
+         
+         loader.loadRecords(itr);
          batcher.flushAndWait();
          moveMgr.stopJob(ticket);
+         
+         ObjectNode beforeDocRoot = objectMapper.createObjectNode();
+         beforeDocRoot.put("job id", id);
+         beforeDocRoot.put("reconciliation",  reconciliation);
+         beforeDocRoot.put("Date",  dateTime.toString());
+         beforeDocRoot.put("headerValue",  headerValue);
```
ehennum: The goal is to add a headerValues property whose value is a JSON array whose items are the strings from the field names on line 110.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
+         ObjectNode csvNode = itr.next();
+         
+         Iterator<String> csvHeader = csvNode.fieldNames();
+         String headerValue = csvNode.get(csvHeader.next()).toString();
+         
+   // TODO: write before job document in /jobStart and /jobs/ID collections or send before job payload to DHF endpoint        
+         String beforeDocId = "/jobs/"+id+"/beforeJob.json";
+         
+         DocumentMetadataHandle beforeDocumentMetadata = new DocumentMetadataHandle();
+         beforeDocumentMetadata.withCollections("/jobs/"+beforeDocId, "/beforeJob");
```
ehennum: By the end of the sprint, it will be necessary to specify permissions on the document metadata handles, but that can happen in a subsequent commit.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
+         loader.loadRecord(csvNode);
+         
+         loader.loadRecords(itr);
          batcher.flushAndWait();
          moveMgr.stopJob(ticket);
+         
+         ObjectNode beforeDocRoot = objectMapper.createObjectNode();
+         beforeDocRoot.put("job id", id);
+         beforeDocRoot.put("reconciliation",  reconciliation);
+         beforeDocRoot.put("Date",  dateTime.toString());
```
ehennum: Isn't dateTime already a string?

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -64,20 +95,42 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
-         loader.loadRecords(converter.convertObject(csvRecords));
+         loader.loadRecord(csvNode);
+         
+         loader.loadRecords(itr);
          batcher.flushAndWait();
          moveMgr.stopJob(ticket);
+         
+         ObjectNode beforeDocRoot = objectMapper.createObjectNode();
+         beforeDocRoot.put("job id", id);
+         beforeDocRoot.put("reconciliation",  reconciliation);
```
ehennum: Instead, let's add a metadata key whose value is the job metadata node. 

Apologies for the muddy guidance on this point.

anu3990:
Its was merged by mistake but I have reverted it back.

anu3990:
Opened a new pull request - https://github.com/marklogic/java-client-api/pull/1027 which includes the changes of this pull request and the after job document.

PR #1031, opened 2018-11-16
Tasks 1469 and 1504.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/DHFPROD-1504 and 
https://project.marklogic.com/jira/browse/DHFPROD-1469
* Are you modifying the correct branch? develop - 1275
* OS and version

### DIFF
--- a/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
+++ b/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
@@ -11,17 +11,66 @@ import com.marklogic.client.datamovement.JobTicket;
 import com.marklogic.client.datamovement.WriteBatcher;
 import com.marklogic.client.document.JSONDocumentManager;
 import com.marklogic.client.io.DocumentMetadataHandle;
+import com.marklogic.client.io.DocumentMetadataHandle.Capability;
 import com.marklogic.client.io.JacksonHandle;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -129,10 +216,15 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
          String ingestionStopTime = LocalDateTime.now().toString();
```
ehennum: Should we log the total number of records that were written to the database after the stopJob() call?

Also, should we collect the stop time after writing the after document?

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/ObjectLoader.java
```diff
@@ -34,6 +37,8 @@ public void loadRecords(Iterator<ObjectNode> objectItr) {
    public void loadRecord(ObjectNode record) {
       String uri = directory+UUID.randomUUID().toString()+".json";
       count++;
+      record.set("instance", this.recordDef);
+      record.set("metadata", this.recordMetadata);
```
ehennum: Let's create a new ObjectNode after line 39 and set the "instance" property of the new object node to record and the "metadata" property of the new object node to recordMetadata.  That way, the new ObjectNode becomes a wrapper for the record converted from a CSV row and the metadata provided with the job.

anu3990:
Included all the suggested changes.

PR #1040, opened 2018-12-06
Initial Commit for Sprint 2 (Sub-tasks)

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-14 and https://project.marklogic.com/jira/browse/JAVA-15
* OS and version

### DIFF
--- a/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
+++ b/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
@@ -148,7 +148,7 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
 				}).onBatchFailure((batch, throwable) -> {
 					if (getBatchSize() == 1) {
 						logger.info("Failure for the batch-" + batch.getJobBatchNumber() + ". Total failures - 1");
-						jobRunnerVariables.lastFailureCount.set(jobRunnerVariables.totalSuccessCount.longValue() + 1);
+						jobRunnerVariables.lastFailureCount.set(jobRunnerVariables.totalFailureCount.longValue() + 1);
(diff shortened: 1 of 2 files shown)


anu3990:
Removed the one-batch size code and moved the "jobRunnerVariables.totalSuccessCount.incrementAndGet();" before threshold check for logging. So now we do not have to handle the plus one while logging.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -124,49 +124,33 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
 				batcher = batcher.onBatchSuccess(batch -> {
-					if (getBatchSize() == 1) {
-						logger.info("Success for the batch-" + batch.getJobBatchNumber() + ". Total successes - 1");
-						jobRunnerVariables.lastSuccessCount.set(jobRunnerVariables.totalSuccessCount.longValue() + 1);
-						jobRunnerVariables.lastSuccessLogTime.set(System.currentTimeMillis());
-					} else {
 						long lastTime = jobRunnerVariables.lastSuccessLogTime.longValue();
 						long currentTime = System.currentTimeMillis();
 						long threshhold = lastTime + (getTimeIntervalInSeconds() * 1000);
+						jobRunnerVariables.totalSuccessCount.incrementAndGet();
```
ehennum: Good idea.  
For that matter, how about capturing the result of incrementAndGet() by setting the currentTotalSuccessCount variable outside the branch instead of inside the branch?  That way, no separate get operation is required, so no other thread has an opportunity to change the value.

anu3990:
Yes that's correct. 
Moving currentTotalSuccessCount and currentTotalFailureCount assignments outside the branches.

PR #1042, opened 2018-12-13
Sprint 2 - Java 12 and 16 (Also corrected the indentation). Continuation from pull request - 1041 

anu3990:
https://project.marklogic.com/jira/browse/JAVA-12 and
https://project.marklogic.com/jira/browse/JAVA-16
* Are you modifying the correct branch? develop-1275
* OS and version

### DIFF
--- a/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CSVConverter.java
+++ b/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CSVConverter.java
@@ -29,11 +29,11 @@ public class CSVConverter {
 
 // TODO: in job control file, expose schema control over whether the value is a JSON string, number, or boolean
 	public Iterator<ObjectNode> convertObject(InputStream csvSource) throws IOException {
-		CountInputStream countInputStream = new CountInputStream(csvSource);
+		this.countInputStream =  new CountInputStream(csvSource);
 		CsvSchema firstLineSchema = CsvSchema.emptySchema().withHeader();
 
 		Iterator<ObjectNode> objectItr = csvMapper.readerFor(ObjectNode.class).with(firstLineSchema)
-				.readValues(countInputStream);
+				.readValues(this.countInputStream);
 		return objectItr;
 	}
 
--- a/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CSVConverter.java
+++ b/ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CSVConverter.java
@@ -1,7 +1,6 @@
 package com.marklogic.client.dhs;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
-// import com.fasterxml.jackson.dataformat.csv.CsvFactory;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvParser;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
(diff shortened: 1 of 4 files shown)


ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CSVConverter.java
```diff
@@ -11,52 +10,31 @@
...
-      CsvParser parser = factory.createParser(csvSource);
-      parser.setSchema(firstLineSchema);
-      Iterator<String> stringItr = parser.readValuesAs(String.class);
-
-      return stringItr;
-   }
- */
-}
+	public Iterator<ObjectNode> convertObject(InputStream csvSource) throws IOException {
+		CountInputStream countInputStream = new CountInputStream(csvSource);
```
ehennum: Should this line assign the field instead of a local variable so the CSVConverter.getByteCount() method calls this instance of CountInputStream?

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CountInputStream.java
```diff
@@ -0,0 +1,43 @@
...
+			byteCount += readLength;
+		return readLength;
+	}
+
+	@Override
+	public int read() throws IOException {
+		int readLength = super.read();
+		if (readLength > 0)
+			byteCount += readLength;
+		return readLength;
```
ehennum: Could this method be implemented with
`return addToByteCount(super.read());`

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/JobRunner.java
```diff
@@ -92,162 +91,210 @@ public void run(DatabaseClient client, InputStream csvRecords, InputStream jobCo
...
-							jobRunnerVariables.lastSuccessCount.set(currentTotalSuccessCount);
+
+					long lastTime = jobRunnerVariables.lastSuccessLogTime.longValue();
+					long currentTime = System.currentTimeMillis();
+					long threshhold = lastTime + (getTimeIntervalInSeconds() * 1000);
+					long currentTotalSuccessCount = jobRunnerVariables.totalSuccessCount.incrementAndGet();
+					if (threshhold <= currentTime
+							&& jobRunnerVariables.lastSuccessLogTime.compareAndSet(lastTime, currentTime)) {
+						ObjectNode statusNode = new ObjectMapper().createObjectNode();
+						statusNode.put("Timestamp", new Timestamp(System.currentTimeMillis()).toString());
```
ehennum: Could the status object have a "Current status" property similar to all of the other status messages?

ml-dhs-aws/src/test/java/com/marklogic/client/dhs/test/TestJobRunner.java
```diff
@@ -30,174 +28,204 @@
...
+	public class TestJobLogger implements JobLogger {
+		int countSuccesses = 0;
+		int countFailures = 0;
+
+		@Override
+		public void status(ObjectNode obj) {
+			countSuccesses++;
+			assertNotNull(obj);
+			assertNotNull(obj.path("Timestamp"));
+			assertNotNull(obj.path("Current status"));
```
ehennum: Where the "Current status" property indicates that we're logging batch success, could the test verify that the expected properties for a batch status message are not null?  The motivation would be to confirm that the status logging is providing what a progress indicator would need.

anu3990:
Added the changes. Also modified the test case a bit. 
Yes sure, I can add System.out.println()  for the demo.

ml-dhs-aws/src/main/java/com/marklogic/client/dhs/CountInputStream.java
```diff
@@ -28,7 +28,7 @@ public int read() throws IOException {
 		int readLength = super.read();
 		if (readLength > 0)
 			byteCount += readLength;
-		return readLength;
+		return addToByteCount(super.read());
```
ehennum: The suggestion was to delete lines 28 - 30 as well.  At present, the revision reads two bytes and counts two bytes.

PR #1048, opened 2019-01-11
Test branch

anu3990:
https://project.marklogic.com/jira/browse/JAVA-19,
https://project.marklogic.com/jira/browse/JAVA-20,
https://project.marklogic.com/jira/browse/JAVA-21 and 
https://project.marklogic.com/jira/browse/JAVA-22
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
@@ -523,10 +523,10 @@ public class DatabaseClientFactory {
   
 	public static class SAMLAuthContext implements SecurityContext {
 		private String token;
-		SSLContext sslContext;
-		X509TrustManager trustManager;
-		SSLHostnameVerifier sslVerifier;
-		Function<ExpiringSAMLAuth, ExpiringSAMLAuth> authorizer;
+		private SSLContext sslContext;
+		private X509TrustManager trustManager;
+		private SSLHostnameVerifier sslVerifier;
+		private Function<ExpiringSAMLAuth, ExpiringSAMLAuth> authorizer;
 		
 		public X509TrustManager getTrustManager() {
 			return trustManager;
@@ -543,23 +543,24 @@ public class DatabaseClientFactory {
 		}
 
 		public interface ExpiringSAMLAuth {
-	        public static ExpiringSAMLAuth newExpiringSAMLAuth(final String authorizationToken, final Instant expiry) {
-	            return new ExpiringSAMLAuth() {
-					
-					@Override
-					public Instant getExpiry() {
-						return expiry;
-					}
-					
-					@Override
-					public String getAuthorizationToken() {
-						return authorizationToken;
-					}
-				};
-	        }
 	        public String getAuthorizationToken();
 	        public Instant getExpiry();
 	    }
+		
+		public static ExpiringSAMLAuth newExpiringSAMLAuth(final String authorizationToken, final Instant expiry) {
+            return new ExpiringSAMLAuth() {
+                
+                @Override
+                public Instant getExpiry() {
+                    return expiry;
+                }
+                
+                @Override
+                public String getAuthorizationToken() {
+                    return authorizationToken;
+                }
+            };
+        }
 
 		@Override
 		public SAMLAuthContext withSSLContext(SSLContext context, X509TrustManager trustManager) {
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
@@ -544,9 +544,19 @@ public class DatabaseClientFactory {
 		public Function<ExpiringSAMLAuth, ExpiringSAMLAuth> getAuthorizer() {
             return authorizer;
         }
-
+		
+		/**
+		 * ExpiringSAMLAuth is used by SAMLAuthContext for reauthorization.
+		 */
 		public interface ExpiringSAMLAuth {
+		    /**
+		     * @return a new SAML assertion token.
+		     */
 	        public String getAuthorizationToken();
+	        
+	        /**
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -0,0 +1,66 @@
...
+	    }
+	 
+	    String samlHeaderValue = RESTServices.AUTHORIZATION_TYPE_SAML+ " "+ RESTServices.AUTHORIZATION_PARAM_TOKEN + "=" + authorizationTokenValue;
+	    Request authenticatedRequest = request.newBuilder().header(RESTServices.HEADER_AUTHORIZATION, samlHeaderValue).build();
+	    return chain.proceed(authenticatedRequest);
+	  }
+	  
+	  private synchronized void authorize(Instant expiry) {
+		  if(expiringSAMLAuth == null)
+			  throw new MarkLogicInternalException("ExpiringSAMLAuth cannot be null");
```
ehennum: On the first call, both the expiry param and the expiringSAMLAuth field will be null.  

So, I believe this test will cause the first call to fail.

Instead, if the expiry param is null but the expiringSAMLAuth field is not null, some other thread must have updated the expiringSAMLAuth field, so the authorize() method should return.

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -0,0 +1,66 @@
...
+		  expiringSAMLAuth = authorizer.apply(expiringSAMLAuth);
+		  if(expiringSAMLAuth==null)
+			  throw new IllegalArgumentException("SAML Authentication cannot be null");
+		  
+		  if(expiringSAMLAuth.getAuthorizationToken() == null)
+			  throw new IllegalArgumentException("SAML Authentication token cannot be null");
+		  authorizationTokenValue = expiringSAMLAuth.getAuthorizationToken();
+		  
+		  if(expiringSAMLAuth.getExpiry() == null)
+			  throw new IllegalArgumentException("SAML authentication does not have expiry value.");
```
ehennum: One other validation occurs to partially addressing the issue that you raised in our meeting yesterday -- if the expiry instant .isBefore(Instant.now), the token has already expired.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -238,45 +237,138 @@ private FailedRequest extractErrorFields(Response response) {
...
-
+  public void connect(String host, int port, String database, SecurityContext securityContext){
+	String user = null;
+	Map<String, String> kerberosOptions = null;
+	String password = null;
+	Authentication type = null;
+	SSLContext sslContext = null;
+	SSLHostnameVerifier sslVerifier = null;
+	X509TrustManager trustManager = null;
+	String authorizationTokenValue = null;
```
ehennum: Can we remove some of these variables if we move the contents of the branches into methods?

Would the code be easier to read and thus understand if the indentation of the variables that remain aligned with line 249?

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -238,45 +237,138 @@ private FailedRequest extractErrorFields(Response response) {
...
+			if (basicContext.getTrustManager() != null)
+				trustManager = basicContext.getTrustManager();
+			if (basicContext.getSSLHostnameVerifier() != null) {
+				sslVerifier = basicContext.getSSLHostnameVerifier();
+			} else {
+				sslVerifier = SSLHostnameVerifier.COMMON;
+			}
+			interceptor = new HTTPBasicAuthInterceptor(credentials);
+	        checkFirstRequest = false;
+		}
```
ehennum: If I recall correctly, an earlier version moved the SSL setup out of the branches and moved the other code from the branches into methods called from the branch by casting the security context and passing the client builder. 

Would that approach still work?

anu3990: Included configureAuthentication overloads. Sorry it was missed during the branch change.

marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -509,6 +520,88 @@ public KerberosAuthContext withSSLHostnameVerifier(SSLHostnameVerifier verifier)
...
+					public Instant getExpiry() {
+						return expiry;
+					}
+					
+					@Override
+					public String getAuthorizationToken() {
+						return authorizationToken;
+					}
+				};
+	        }
```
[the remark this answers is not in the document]
anu3990: Yes that is correct. Including the change.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
ehennum: Would it work to pass the SecurityContext and OkHttpClient.Builder to each configureAuthentication() method?

As in

    if (securityContext instanceof BasicAuthContext) {
        clientBldr = configureAuthentication((BasicAuthContext) securityContext, clientBldr);
    } else  (securityContext instanceof DigestAuthContext) {
        clientBldr = configureAuthentication((DigestAuthContext) securityContext, clientBldr);
    ... other tests and casts on calls ...

In other words, to move the code that accesses the security context into the configureAuthentication() methods?

Also, given that the SSL getters are defined on SecurityContext, could the SSL configuration be performed once outside of the branches on the specific instances of SecurityContext?

Is that possible?

anu3990:
Added copyright and javadoc corresponding to https://project.marklogic.com/jira/browse/JAVA-27 and https://project.marklogic.com/jira/browse/JAVA-28

PR #1049, opened 2019-01-18
call a renewer when needed.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-29
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
@@ -36,7 +36,7 @@ public class HTTPSamlAuthInterceptor implements Interceptor {
     private ExpiringSAMLAuth expiringSAMLAuth;
     private long threshold;
     private RenewerCallback renewer;
-    private AtomicBoolean isCallbackExecuting = new AtomicBoolean(false);
+    private AtomicBoolean isCallbackExecuting;
 
     public HTTPSamlAuthInterceptor(String authToken) {
         this.authorizationTokenValue = authToken;
@@ -49,23 +49,21 @@ public class HTTPSamlAuthInterceptor implements Interceptor {
     public HTTPSamlAuthInterceptor(ExpiringSAMLAuth authorization, RenewerCallback renew) {
         expiringSAMLAuth = authorization;
         renewer = renew;
+        isCallbackExecuting = new AtomicBoolean(false);
     }
 
     @Override
     public Response intercept(Chain chain) throws IOException {
         Request request = chain.request();
         if (authorizer != null) {
-            if ((expiringSAMLAuth == null || threshold <= Instant.now().getEpochSecond())
-                    && isCallbackExecuting.compareAndSet(false, true)) {
+            if ((expiringSAMLAuth == null || threshold <= Instant.now().getEpochSecond())) {
                 authorizeCallback();
             }
-        } else if (renewer != null) {
-            if (threshold <= Instant.now().getEpochSecond()) {
-                RenewExpiry renewExpiry = new RenewExpiry();
+        } else if (renewer != null && threshold <= Instant.now().getEpochSecond()) {
+                RenewCallback renewExpiry = new RenewCallback(expiringSAMLAuth);
                 if (isCallbackExecuting.compareAndSet(false, true))
                     Executors.defaultThreadFactory().newThread(renewExpiry).start();
 
-            }
         }
         String samlHeaderValue = RESTServices.AUTHORIZATION_TYPE_SAML + " " + RESTServices.AUTHORIZATION_PARAM_TOKEN
                 + "=" + authorizationTokenValue;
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -29,96 +29,98 @@
...
+        Request request = chain.request();
+        if (authorizer != null) {
+            if ((expiringSAMLAuth == null || threshold <= Instant.now().getEpochSecond())
+                    && isCallbackExecuting.compareAndSet(false, true)) {
+                authorizeCallback();
+            }
+        } else if (renewer != null) {
+            if (threshold <= Instant.now().getEpochSecond()) {
+                RenewExpiry renewExpiry = new RenewExpiry();
+                if (isCallbackExecuting.compareAndSet(false, true))
```
ehennum: Would it be slightly cleaner and more efficient if the compareAndSet() test were anded to the threshold test on line 63?

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -29,96 +29,98 @@
...
+            }
+            authorizationTokenValue = expiringSAMLAuth.getAuthorizationToken();
+            checkAuthorizationExpiry(expiringSAMLAuth.getExpiry());
+        } finally {
+            isCallbackExecuting.set(false);
+        }
+
+    }
+
+    private void checkAuthorizationExpiry(Instant instant) {
```
ehennum: Would the name setThreshold(Instant expiry) better convey what the method is doing?

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -29,96 +29,98 @@
...
+        } finally {
+            isCallbackExecuting.set(false);
+        }
+    }
+
+
+    private class RenewExpiry implements Runnable {
+        @Override
+        public void run() {
+            renewCallback(expiringSAMLAuth);
```
ehennum: Should the run() method pass the returned Instant to setThreshold(Instant expiry)?

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -29,96 +29,98 @@
...
-	  }
     @Override
-    public void run() {
-        renew(expiringSAMLAuth);
-        
+    public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        if (authorizer != null) {
+            if ((expiringSAMLAuth == null || threshold <= Instant.now().getEpochSecond())
+                    && isCallbackExecuting.compareAndSet(false, true)) {
```
ehennum: Apologies, but it sank in that using compareAndSet() for the authorizer was a bad suggestion.

The problem is that we need to block all threads until the authorizer gets the new authorization token.

In other words, the previous synchronized approach for the authorizer was better.

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -29,96 +29,98 @@
...
-	  public HTTPSamlAuthInterceptor(ExpiringSAMLAuth authorization, RenewerCallback renew) {
-	    expiringSAMLAuth = authorization;
+public class HTTPSamlAuthInterceptor implements Interceptor { 
+                                                                       
+    private String authorizationTokenValue;
+    private AuthorizerCallback authorizer;
+    private ExpiringSAMLAuth expiringSAMLAuth;
+    private long threshold;
+    private RenewerCallback renewer;
+    private AtomicBoolean isCallbackExecuting = new AtomicBoolean(false);
```
ehennum: Given the belated realization that the compareAndSet() can't be used for the authorizer (see below), maybe isCallbackExecuting should be initialized only in the constructor that takes the renewer.

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
ehennum: That is, would it work to delete line 64 and change line 62 as follows:

    } else if (renewer != null && threshold <= Instant.now().getEpochSecond() &&
         isCallbackExecuting.compareAndSet(false, true)) {

marklogic-client-api/src/main/java/com/marklogic/client/impl/HTTPSamlAuthInterceptor.java
```diff
@@ -74,26 +72,22 @@ public Response intercept(Chain chain) throws IOException {
...
     }
 
-    private void authorizeCallback() {
-        try {
-            expiringSAMLAuth = authorizer.apply(expiringSAMLAuth);
-
-            if (expiringSAMLAuth == null) {
-                throw new IllegalArgumentException("SAML Authentication cannot be null");
-            }
+    private synchronized void authorizeCallback() {
```
ehennum: To eliminate the possibility of a race condition, please restore the old expiry parameter and the testing of the old expiry from the version of this method prior to the isCallbackExecuting revision.

PR #1052, opened 2019-01-28
Technical: Splitter with CSV Implementation

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-26
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
@@ -68,7 +68,7 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
                 .configure(CsvParser.Feature.SKIP_EMPTY_LINES, true)
                 .configure(CsvParser.Feature.TRIM_SPACES, true)
                 .configure(CsvParser.Feature.WRAP_AS_ARRAY, false)
-                .configure(CsvParser.Feature.IGNORE_TRAILING_UNMAPPABLE, false);
+                .configure(CsvParser.Feature.IGNORE_TRAILING_UNMAPPABLE, true);
         }
         return csvMapper;
     }
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.marklogic.client.datamovement;
 
 import java.io.InputStream;
@@ -51,7 +67,8 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
                 .configure(CsvParser.Feature.INSERT_NULLS_FOR_MISSING_COLUMNS, false)
                 .configure(CsvParser.Feature.SKIP_EMPTY_LINES, true)
                 .configure(CsvParser.Feature.TRIM_SPACES, true)
-                .configure(CsvParser.Feature.WRAP_AS_ARRAY, false);
+                .configure(CsvParser.Feature.WRAP_AS_ARRAY, false)
+                .configure(CsvParser.Feature.IGNORE_TRAILING_UNMAPPABLE, false);
         }
         return csvMapper;
     }
@@ -64,7 +81,7 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
         }
         Iterator<JsonNode> nodeItr = configureObjReader().readValues(input);
         return StreamSupport.stream(Spliterators.spliteratorUnknownSize(nodeItr, Spliterator.ORDERED), false)
-                .map(this::configureJacksonHandle);
+                .map(this::wrapJacksonHandle);
     }
     public Stream<JacksonHandle> split(Reader input) throws Exception  { 
 
@@ -73,7 +90,7 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
         }
         Iterator<JsonNode> nodeItr = configureObjReader().readValues(input);
         return StreamSupport.stream(Spliterators.spliteratorUnknownSize(nodeItr, Spliterator.ORDERED), false)
-                .map(this::configureJacksonHandle);
+                .map(this::wrapJacksonHandle);
     }
 
     @Override
@@ -88,13 +105,13 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
     private ObjectReader configureObjReader() {
         this.count=0;
         CsvSchema firstLineSchema = getCsvSchema()!=null? getCsvSchema():CsvSchema.emptySchema().withHeader();
-        
-        ObjectReader objectReader = configureCsvMapper().readerFor(JsonNode.class);
+        CsvMapper csvMapper = getCsvMapper()!=null ? getCsvMapper() : configureCsvMapper();
+        ObjectReader objectReader = csvMapper.readerFor(JsonNode.class);
         
         return objectReader.with(firstLineSchema);
     }
     
-    private JacksonHandle configureJacksonHandle(JsonNode content) {
+    private JacksonHandle wrapJacksonHandle(JsonNode content) {
         incrementCount();
         return new JacksonHandle(content);
     }
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
@@ -22,21 +22,28 @@ import com.marklogic.client.io.JacksonHandle;
 public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
     private CsvSchema csvSchema = null;
     private CsvMapper csvMapper;
-    private ObjectReader objectParser;
-    private long count;
+    private long count = 0;
     
-    public JacksonCSVSplitter() {
-        setCsvMapper();
-        setCount(0);
+    public CsvMapper getCsvMapper() {
+        return csvMapper;
+    }
+
+    public JacksonCSVSplitter withCsvSchema(CsvSchema schema) {
+        this.csvSchema = schema;
+        return this;
+    } 
+    
+    public JacksonCSVSplitter withCsvMapper(CsvMapper mapper) {
+        this.csvMapper = mapper;
+        return this;
     }
     
-    public JacksonCSVSplitter(CsvSchema newSchema) {
-        setCsvSchema(newSchema);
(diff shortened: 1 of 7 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -0,0 +1,94 @@
...
+        setCount(0);
+    }
+    
+    public JacksonCSVSplitter(CsvSchema newSchema) {
+        setCsvSchema(newSchema);
+        setCsvMapper();
+        setCount(0);
+    }
+    
+    private void setCsvMapper() {
```
ehennum: On looking at the concrete refactoring, I'm concluding that the Technical Story has a design error.

We always want to convert CSV to JsonNodes, so we should expose ObjectReader.

Instead, we should expose CsvMapper.  That is, instead of having a setCsvMapper() that initializes a private field, we should have a private factory that constructs a default mapper only if the user didn't provide a CsvMapper.  

Then, in the configureObjReader() private factory, we can construct a private ObjectReader for JsonNode objects.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -0,0 +1,94 @@
...
+    public CsvSchema getCsvSchema() {
+        return csvSchema;
+    }
+
+    @Override
+    public Stream<JacksonHandle> split(InputStream input) throws Exception { 
+
+        Iterator<JsonNode> nodeItr = configureObjReader().readValues(input);
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(nodeItr, Spliterator.ORDERED), false)
+                .map(JacksonHandle::new);
```
ehennum: The mapper also need to increment the count.  

So, let's add a private mapper method that takes a JsonNode, increments the count, and constructs and returns a new JacksonHandle for the JsonNode.  Then, we can pass a reference to the mapper here and on line 74.  Or, better, we can pass the Iterator<JsonNode> into a common private method.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public long getCount() { 
+        return this.count;
+    }
+    public void incrementCount() {
```
ehennum: Is there a rationale for exposing a public method to increment the count?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/WriteBatcherImpl.java
```diff
@@ -222,6 +225,7 @@
   private JobTicket jobTicket;
   private Calendar jobStartTime;
   private Calendar jobEndTime;
+  private DocumentMetadataHandle docMetadataHandle;
```
ehennum: Instead of using the type name for the field, could we use a name that suggests the use of the field?  Perhaps defaultMetadata?

That way, it's easier to understand for someone editing the code later (including oneself).

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+    public void setCsvSchema(CsvSchema schema) {
+        this.csvSchema = schema;
+    } 
+    public CsvSchema getCsvSchema() {
+        return csvSchema;
+    }
+
+    @Override
+    public Stream<JacksonHandle> split(InputStream input) throws Exception {
```
ehennum: Would it be appropriate to throw an IllegalArgumentException if the input is null (both for the InputStream and the Reader below)?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/WriteBatcherImpl.java
```diff
@@ -1331,4 +1335,20 @@ public boolean awaitCompletion(long timeout, TimeUnit unit) throws InterruptedEx
...
+
+
+@Override
+public WriteBatcher withDefaultMetadata(DocumentMetadataHandle handle) {
+    this.docMetadataHandle = handle;
+    return this;
+}
+
+@Override
+public void addAll(Stream<? extends DocumentWriteOperation> operations, JacksonCSVSplitter splitter) {
```
ehennum: The addAll() method should only take the first parameter.  The stream of operations could be generated by any of many different sources include other kinds of splitters that we add in the future.

I recognize that the goal was to increment the count, but that's better done by the mapping operation within the splitter (so the two classes are loosely coupled).

That also makes it possible to simplify the iteration by calling the forEach() stream method, passing a reference to the add method. That would eliminate the need to construct an iterator.

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -85,4 +90,88 @@
...
+    public static Stream<DocumentWriteOperation> from(Stream<? extends AbstractWriteHandle> content,
+            final DocumentUriMaker uriMaker) {
+        final class DocumentWriteOperationImpl implements DocumentWriteOperation {
+            public DocumentWriteOperationImpl(AbstractWriteHandle content, String uri) {
+                this.content = content;
+                this.uri = uri;
+            }
+
+            private AbstractWriteHandle content;
+            private String uri;
```
ehennum: For consistency and readability, should the fields move to the top of the class (that is, before the constructor)?

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -85,4 +90,88 @@
...
   String getTemporalDocumentURI();
+  
+  /**
+   * The from method prepares each content object for writing as a document including generating a URI by inserting a UUID.
+   * @param a subclass of AbstractWriteHandle
+   * @param DocumentUriMaker which internally accepts an AbstractWriteHandle and returns a String
+   * @return a stream of DocumentWriteOperation to be written in the database.
+   */
+    public static Stream<DocumentWriteOperation> from(Stream<? extends AbstractWriteHandle> content,
+            final DocumentUriMaker uriMaker) {
```
ehennum: For robustness, should the from() method throw an IllegalArgumentException if either the content or uriMaker is null?

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -85,4 +90,88 @@
...
+        return content.map(wrapperImpl::mapper);
+
+    }
+
+    /**
+     * The uriMaker method creates a uri for each document written in the database
+     * @param format refers to the pattern passed.
+     * @return DocumentUriMaker which contains the formatted uri for the new document.
+     */
+    public static DocumentUriMaker uriMaker(String format) {
```
ehennum: For robustness, should uriMaker() throw an IllegalArgumentException if format is null?

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -85,4 +90,88 @@
...
+
+        final class FormatUriMaker {
+            private String formatter;
```
ehennum: Would the purpose of the field be easier to understand when maintaining the code if the name were format instead of formatter?

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -85,4 +90,88 @@
...
+        final class WrapperImpl {
+            DocumentWriteOperation mapper(AbstractWriteHandle content) {
+                String uri = uriMaker.apply(content);
+                if (uri == null)
+                    throw new MarkLogicInternalException("Uri could not be created");
+                return new DocumentWriteOperationImpl(content, uri);
+            }
+
+        }
+        WrapperImpl wrapperImpl = new WrapperImpl();
```
ehennum: I'm wondering whether a potential improvement on our whiteboarding session would be to pass uriMaker to a WrapperImpl constructor that initialized a private field of WrapperImpl instead of making the uriMaker parameter on the enclosing method final.

That way, WrapperImpl would contain everything needed to make URIs, which might be easier to understand and maintain.  

What do you think?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/WriteBatcherImpl.java
```diff
@@ -1331,4 +1335,20 @@ public boolean awaitCompletion(long timeout, TimeUnit unit) throws InterruptedEx
       return tasks;
     }
   }
+
+
+@Override
+public WriteBatcher withDefaultMetadata(DocumentMetadataHandle handle) {
+    this.docMetadataHandle = handle;
+    return this;
+}
```
ehennum: We'll also need to find the place in the WriteBatcher implementation where the added DocumentWriteOperations are batched for the server and emit the defaultMetadata as the first payload.

Finding the right place may require reading through the implementation and / or stepping through the implementation in a debugger.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -45,50 +52,50 @@ private void setCsvMapper() {
...
-    private void setCount(long countValue) {
-        this.count = countValue;
-    }
     
     private ObjectReader configureObjReader() {
+        this.count=0;
         CsvSchema firstLineSchema = getCsvSchema()!=null? getCsvSchema():CsvSchema.emptySchema().withHeader();
-        ObjectReader objectReader = getObjectReader()!=null? getObjectReader():csvMapper.readerFor(JsonNode.class);
+        
+        ObjectReader objectReader = configureCsvMapper().readerFor(JsonNode.class);
```
ehennum: Similar to the CsvSchema, could we let users override the default?  as in:

```
CsvMapper csvMapper = getCsvMapper()!=null ? CsvMapper() : configureCsvMapper();
ObjectReader objectReader = csvMapper.readerFor(JsonNode.class);
```

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -45,50 +52,50 @@ private void setCsvMapper() {
...
         
         return objectReader.with(firstLineSchema);
     }
+    
+    private JacksonHandle configureJacksonHandle(JsonNode content) {
```
ehennum: Does "configure" imply initialization?  Where, instead, this operation happens during execution.

Would a name like "wrap()" or "prepare()" be more informative to future maintainers (including ourselves)?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/WriteBatcherImpl.java
```diff
@@ -1339,16 +1341,13 @@ public boolean awaitCompletion(long timeout, TimeUnit unit) throws InterruptedEx
...
 
 @Override
-public void addAll(Stream<? extends DocumentWriteOperation> operations, JacksonCSVSplitter splitter) {
+public void addAll(Stream<? extends DocumentWriteOperation> operations) {
     Iterator<? extends DocumentWriteOperation> docWriteItr = operations.iterator();
-    while(docWriteItr.hasNext()) {
-      add(docWriteItr.next());
-      splitter.incrementCount();
-    }
+    docWriteItr.forEachRemaining(this::add);
```
ehennum: Would it work to drop the iterator and just execute

`    operations.forEach(this::add);`

ehennum:
@georgeajit , what remains to be done on the merge?

PR #1058, opened 2019-02-13
refactor DHS S3CSV component to use CSV splitter.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-35
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
@@ -25,10 +25,13 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -115,4 +130,29 @@ private JacksonHandle wrapJacksonHandle(JsonNode content) {
...
+        csvHeaderNode = headerNode;
+        Iterator<String> headerValue = headerNode.fieldNames();
+        this.headers = new ObjectMapper().createArrayNode();
+        while (headerValue.hasNext()) {
+            headers.add(headerValue.next());
+        }
+    }
+    
+    private PeekingIterator<JsonNode> configureSplitObj(Iterator<JsonNode> nodeItr){
+        if (!nodeItr.hasNext()) {
```
ehennum: To defend against an NPE, should the test be on nodeItr == null || !nodeItr.hasNext()?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -39,7 +42,13 @@
     private CsvSchema csvSchema = null;
     private CsvMapper csvMapper;
     private long count = 0;
+    private ArrayNode headers = null;
+    private JsonNode csvHeaderNode;
```
ehennum: Is there a known use of csvHeaderNode?  If not, it might be best to remove the field and getter so as to maximize encapsulation.

PR #1062, opened 2019-02-21
Adding javadoc.

anu3990:
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
@@ -77,6 +77,12 @@ public class JacksonCSVSplitter implements Splitter<JacksonHandle> {
         return csvMapper;
     }
 
+    /*
+     * Takes the input stream and converts it into a stream of JacksonHandle by setting the schema 
+     *  and wrapping the JsonNode into JacksonHandle.
+     * @param input the input stream passed in.
+     * @return a stream of JacksonHandle.
+     */
     @Override
     public Stream<JacksonHandle> split(InputStream input) throws Exception { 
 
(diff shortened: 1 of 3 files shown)


anu3990:
Please let me know if any file was missed.

anu3990:
Added the missing javadoc in JacksonCSVSplitter and Splitter.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JacksonCSVSplitter.java
```diff
@@ -44,20 +44,36 @@
     private long count = 0;
     private ArrayNode headers = null;
     
+    /**
```
ehennum: My understanding of standard JavaDoc practice is to provide the sentence describing the method as well as the description of the return value.

Or, when you generate the JavaDoc from the sources, does it look okay?

The main concern is that we want it to be intelligible for readers.

The same question applies to other getters here and in Splitter that have a return description but not a method description.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/Splitter.java
```diff
@@ -1,13 +1,42 @@
...
+
 package com.marklogic.client.datamovement;
 
 import java.io.InputStream;
 import java.util.stream.Stream;
 
 import com.marklogic.client.io.marker.AbstractWriteHandle;
 
+/**
+ * To facilitate CSV splitting, Splitter allows to implement the split method where the
```
ehennum: Please describe Splitter as the abstract definition of a class that can split an input stream into a Java stream of write handles.  Typically, the handles are used for writing documents to the database.

(In other words, the abstraction shouldn't mention the concrete implementation -- that's why it's an abstraction.)

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/Splitter.java
```diff
@@ -1,13 +1,42 @@
...
+/**
+ * To facilitate CSV splitting, Splitter allows to implement the split method where the 
+ *  incoming input stream is converted to a stream of another object, for example JacksonHandle 
+ *  and gives an option to keep a record of the number of objects created using getCount method.
+ */
 public interface Splitter<T extends AbstractWriteHandle> {
+    /**
+     * Converts the incoming input stream to a stream of object T.
+     * @param input is the incoming input stream.
+     * @return a stream of T.
```
ehennum: Would it be more useful to the reader to say "a stream of AbstractWriteHandle objects" to the reader doesn't have to look for the meaning of T?

Issue #1055, opened 2019-02-06
FileHandle is not streaming the response to a File


rjrudin:
Found this while analyzing something else with @ehennum - I believe it's expected that when using a FileHandle with something like RowManager.resultDoc, the response from ML should be streamed to the given File. Instead, the entire response is being read in as a String and then written to the File, which is much more expensive. 

@ehennum confirmed that this is not specific to RowManager either, it's generic to FileHandle - com/marklogic/client/impl/OkHttpServices.java#5364-5373

georgeajit:
Verified on develop branch with ML 10.0.1 server build.

PR #1064, opened 2019-03-07
Improving the FileHandle efficiency.

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1055
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/Utilities.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/Utilities.java
@@ -25,13 +25,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
-import java.io.RandomAccessFile;
 import java.io.Reader;
 import java.io.StringReader;
-import java.nio.ByteBuffer;
-import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
-import java.nio.file.Path;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/Utilities.java
```diff
@@ -669,4 +675,26 @@ static public DatatypeFactory getDatatypeFactory() {
...
+      while (in.read(buf) != -1) {
+        byteBuf = ByteBuffer.wrap(buf);
+        byteBuf.flip();
+        inChannel.write(byteBuf);
+        byteBuf.clear();
+      }
+      outFile.close();
+      } catch (IOException e) {
+          throw new MarkLogicIOException("Internal Exception occured.");
+      }
```
aebadirad: Might want to add a 'finally { outfile.close() } ' if it exists to just limit the time it takes to close out the file handle.

Also, a large initial direct byte buffer will lessen the impact buffer allocation has if one direct buffer is created for universal use, then 'slices' are taken off to write files then returning the memory when the write occurs in equal chunks. This will reuse native memory outside the java heap at a lower level and should give a noticeable boost if writing binaries that are significantly above the L1+L2 cache size (which binaries tend to be). 

Just a few random thoughts.

anu3990: Thanks Alex.!
So you mean we should have something like - 
byte[] buf = new byte[in.available()];
ByteBuffer byteBuf = ByteBuffer.allocate(in.available());
if(in.read(buf) != -1) {
      byteBuf = ByteBuffer.wrap(buf);
 }
and then write out the byteBuffer in chunks using FileChannel. 
Am i correct.?

marklogic-client-api/src/main/java/com/marklogic/client/impl/Utilities.java
```diff
@@ -676,25 +673,38 @@ static public DatatypeFactory getDatatypeFactory() {
     return datatypeFactory;
   }
   
-  static public void write(InputStream in, Path path) { 
+  static public void write(InputStream in, OutputStream outStream) throws IOException {
```
ehennum: Should this method simply return if in or out is null?

marklogic-client-api/src/main/java/com/marklogic/client/impl/Utilities.java
aebadirad: Nah, meaning DirectByteBuffer (ByteBuffer.allocateDirect()) and reuse it as your bytebuffer throughout (allocate once, reuse multiple). A direct allocation should be significantly faster than a non-direct buffer writing to channels, as the native OS level code doesn't have to keep dipping back into the JVM to read, instead it can pull from the direct buffer. 

Though if we're wanting to go pure performance, @ehennum, do we want to consider off heap for the reads/writes from disk? It's in unsafe space, but so is technically a direct allocation of memory, and would greatly improve the read performance (as well as the write)?

PR #1068, opened 2019-03-18
develop-dsbulk-coreimpl

ehennum:
implementation of bare minimum for a batched call 
in the simplest case


ehennum:
In hopes of simplifying the surface area, I moved some interfaces into static classes.  I also had to push some implementation classes down into an impl subpackage so they could be public and imported in other packages but still undocumented.

Please see whether you think this work is headed in the right direction.  (Much still to do.)

PR #1069, opened 2019-03-21
CallBatcherTest for core functionality of bulk calls

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-47
* Are you modifying the correct branch? develop-dsbulk
* OS and version

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherTest.java
@@ -38,8 +38,11 @@ import com.marklogic.client.test.Common;
 import com.marklogic.client.test.dataservices.CallManagerTest.EndpointSetup;
 
 public class CallBatcherTest {
+	
 	  private final static String ENDPOINT_DIRECTORY = "/javaApi/test/callManager/";
+	  private static EndpointSetup endpointSetup = new EndpointSetup(ENDPOINT_DIRECTORY);
 	  private static ObjectMapper objectMapper = new ObjectMapper();
+	  
 	  private DatabaseClient db      = Common.connect();
 	  private CallManager    callMgr = CallManager.on(db);
 	  private JacksonHandle serviceHandle;
@@ -53,7 +56,7 @@ public class CallBatcherTest {
 	  private CallManager.CallableEndpoint installEndpoint(String functionName) {
 		  if(functionName == null || functionName.length() == 0)
 			  throw new MarkLogicInternalException("Invalid input was sent.");
-		  JsonNode endpointdef = CallManagerTest.endpointdefs.get(functionName);
+		  JsonNode endpointdef = endpointSetup.endpointdefs.get(functionName);
 	      return callMgr.endpoint(serviceHandle, new JacksonHandle(endpointdef), "sjs");
 	  }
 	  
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherTest.java
```diff
@@ -0,0 +1,336 @@
...
+		  batcher.onCallSuccess(event-> { 
+			  assignedParams.add(event.getArgs().getAssignedParamNames()[0]);
+		  });
+		  
+		  batcher.getDataMovementManager().startJob(batcher);
+		  
+	      batcher.add(noneCaller.args().param("param1", 1.2));
+
+	      batcher.flushAndWait();
+	      assertTrue(assignedParams.size()!=0);
```
ehennum: Could the assertion be more precise?  Would the test result be valid if the array had more than 1 item?

Also, could the assertion be more informative if it fails?  As in:

`    assertEqual("invalid number of parameters", assignedParams.size(), 1); 
`
The same comment applies to the other tests.

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherTest.java
```diff
@@ -0,0 +1,336 @@
...
+	  @Test
+	  public void oneForArgsTest() {
+		  CallManager.CallableEndpoint callableEndpoint = installEndpoint("oneParam");
+		  CallManager.OneCaller<Double> oneCaller = callableEndpoint.returningOne(Double.class);
+		  
+		  CallBatcher<CallManager.CallArgs,CallManager.OneCallEvent<Double>> batcher = oneCaller.batcher().forArgs();
+		  assertNotNull(batcher);
+		  ArrayList<String> assignedParams = new ArrayList<String>();
+		  batcher.onCallSuccess(event-> { 
+			  assignedParams.add(event.getArgs().getAssignedParamNames()[0]);
```
ehennum: OneCallEvent<Double> has a Double getItem() method that provides access to the return value.

Could the success event capture that return value in a local variable and confirm it with an assertion? after flushing and waiting so that the test provides more complete coverage of the intended behavior?

The same comment applies to the manyForArgsTest with the difference that ManyCallEvent<Double> has a getItems() method.

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherTest.java
```diff
@@ -0,0 +1,336 @@
...
+	    }
+	    
+	    private static void setupEndpointSingleRequired(
+	            JSONDocumentManager docMgr, DocumentMetadataHandle docMeta, String functionName, String datatype
+	    ) {
+	        JsonNode endpointdef = getEndpointdef(functionName, datatype, false, false);
+	        String script = getScript(datatype,null,null, false, false);
+	        setupEndpoint(docMgr, docMeta, endpointdef, script);
+	    }
+
```
ehennum: For the sake of more maintainability, could the endpoint creation methods in CallManagerTest become utilities that could be called from either CallManagerTest or CallBatcherTest?  

Likely those functions use some fields of CallManagerTest.  Maybe those fields could move into a static nested class of CallManagerTest, the shared functions could become methods of that static nested class, and CallManagerTest and CallBatcherTest could each have an instance of that static nested class that's used for setup.

What do you think?

anu3990: Yes that's right. 
Moving the methods to the static nested class EndpointSetup.

ehennum: The nested class has dependencies on CallManagerTest.

If both CallManagerTest and CallBatcherTest construct an instance of the setup class, could each configure the instance for its needs?  For instance if the constructor is passed in the endpoint directory, could the constructor of the nested setup class create the service definition JsonNode as a field such that the methods (if the methods of the setup class become object methods instead of static methods) can have access to the field?

ehennum:
Thanks for iterating on the changes

Issue #1070, opened 2019-03-26
multiple clients for Data Service bulk calls


ehennum:
Support multiple clients (default to one per host) for Data Service bulk calls.

ehennum:
Fills in some of the initial implementation.

georgeajit:
Closed.

PR #1071, opened 2019-03-26
determine hosts and clients from forest configuration #1070

ehennum:
adds support for multiple clients


ehennum:
Please take a look and see if you have concerns

PR #1079, opened 2019-04-03
 CallBatcherBuilder.withDefaultArgs

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-58
* Are you modifying the correct branch? develop-dsbulk
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
@@ -35,6 +35,7 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class CallBatcherImpl<W, E extends CallManager.CallEvent> extends BatcherImpl implements CallBatcher<W,E> {
@@ -423,14 +424,18 @@ public class CallBatcherImpl<W, E extends CallManager.CallEvent> extends Batcher
 	public CallBatcherImpl<W, E> withdefaultArgs(CallArgs args) {
 		if(args == null)
 			throw new IllegalArgumentException("CallArgs cannot be null.");
+		
 		CallArgsImpl callArgsImpl = (CallArgsImpl) args;
+		
 		if(callArgsImpl == null || ! (callArgsImpl instanceof CallArgsImpl))
 			throw new IllegalArgumentException("Illegal Argument.");
 		if(callArgsImpl.getEndpoint().getEndpointPath()!= caller.getEndpointPath())
 			throw new IllegalArgumentException("Endpoints are different.");
+		if(callArgsImpl.getCallFields() == null || callArgsImpl.getCallFields().size() == 0)
+			return this;
 		
-		List<CallField> callFieldList = callArgsImpl.getCallFields();
-		callFieldList.forEach(p->p.toBuffered());
+		List<CallField> callFieldList = callArgsImpl.getCallFields().stream().map(p->p.toBuffered()).collect(Collectors.toList());
+		
 		this.defaultArgs = callFieldList;
 		return this;
 	}
@@ -455,13 +460,24 @@ public class CallBatcherImpl<W, E extends CallManager.CallEvent> extends Batcher
 		if(defaultArgs == null || defaultArgs.size()==0)
 			return callArgsImpl;
 		CallerImpl callerImpl = (com.marklogic.client.dataservices.impl.CallManagerImpl.CallerImpl) caller;
-		List<CallField> newCallFields = new ArrayList<CallField>();
+		Set<CallField> newCallFields = new HashSet<CallField>();
+		Set<String> assignedParams = new HashSet<String>();
 		
-		if(callArgsImpl.getCallFields()!=null && callArgsImpl.getCallFields().size()!=0)
+		if(callArgsImpl.getCallFields()!=null && callArgsImpl.getCallFields().size()!=0) {
 			newCallFields.addAll(callArgsImpl.getCallFields());
+			assignedParams.addAll(callArgsImpl.getAssignedParams());
+		}
 		
+		for(CallField i: defaultArgs) {
+			if(!assignedParams.contains(i.getParamName())) {
+				assignedParams.add(i.getParamName());
+				newCallFields.add(i);
+			}
+		}
+
 		newCallFields.addAll(defaultArgs);
+		List<CallField> callFieldList = newCallFields.stream().collect(Collectors.toList());
 		
-		return new CallArgsImpl(callerImpl.getEndpoint(), newCallFields);
+		return new CallArgsImpl(callerImpl.getEndpoint(), callFieldList);
 	}
 }
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
@@ -22,7 +22,12 @@ import com.marklogic.client.datamovement.impl.BatcherImpl;
 import com.marklogic.client.dataservices.CallBatcher;
 import com.marklogic.client.dataservices.CallFailureListener;
 import com.marklogic.client.dataservices.CallManager;
+import com.marklogic.client.dataservices.CallManager.CallArgs;
 import com.marklogic.client.dataservices.CallSuccessListener;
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -414,4 +418,50 @@ public Boolean call() {
...
+		if(args == null)
+			throw new IllegalArgumentException("CallArgs cannot be null.");
+		CallArgsImpl callArgsImpl = (CallArgsImpl) args;
+		if(callArgsImpl == null || ! (callArgsImpl instanceof CallArgsImpl))
+			throw new IllegalArgumentException("Illegal Argument.");
+		if(callArgsImpl.getEndpoint().getEndpointPath()!= caller.getEndpointPath())
+			throw new IllegalArgumentException("Endpoints are different.");
+		
+		List<CallField> callFieldList = callArgsImpl.getCallFields();
+		callFieldList.forEach(p->p.toBuffered());
```
ehennum: Does this line store the buffered fields anywhere?

Would it be best to create a new list so the existing list owned by the application CallArgs object is not modified?

If the args have a null or empty field, should the withDefaultArgs() method return this without doing anything?

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -440,6 +440,20 @@ void patchDocument(RequestLogger reqlog, DocumentDescriptor desc, Transaction tr
     public String getParamName() {
       return paramName;
     }
+    public CallField toBuffered() {
+    	return this;
+    }
+ 
+    public interface MultipleAtomicCallField {
```
ehennum: Given that CallField is an abstract class, the story was wrong to suggest that MultipleAtomicCallField, MultipleNodeCallField, and SingleNodeCallField should be interfaces.

Instead, it would be better if these interfaces were static public abstract classes at the same level as CallField that extend CallField to create a proper type hierarchy.

That way, the Buffered and Unbuffered concrete classes can extend the specific abstract class instead of extending CallField but implementing the interface.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -450,37 +464,116 @@ public SingleAtomicCallField(String paramName, String paramValue) {
...
+	    }
+	  
+		@Override
+		public Stream<BytesHandle> getParamValues() {
+			return paramValues;
+		}
+	  
+  }
+  
+  static public class BufferedSingleNodeCallField extends CallField implements SingleNodeCallField {
```
ehennum: A single node call field takes a single node value.  So

- the constructor should take a single BytesHandle instead of a stream
- the private field should store a single BytesHandle
- the single value getter should return the private stream instead of null 
- there should not be a stream getter

Adding @Override marker annotations can help assure that the class is implementing the contract of SingleNodeCallField

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -450,37 +464,116 @@ public SingleAtomicCallField(String paramName, String paramValue) {
...
+  
+	@Override
+	public Stream<String> getParamValues() {
+		return Stream.of(paramValues);
+	}
+  }
+	
+  static public class BufferedMultipleNodeCallField extends CallField implements MultipleNodeCallField {
+	  
+		private Stream<BytesHandle> paramValues;
```
ehennum: As with MultipleAtomicCallField, MultipleNodeCallField should have a private field that is an array.  A stream can only be used once.

In this case, the array would have the BytesHandle type.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -450,37 +464,116 @@ public SingleAtomicCallField(String paramName, String paramValue) {
...
+		public BufferedMultipleNodeCallField(String paramName, Stream<BytesHandle> paramValues) {
+			super(paramName);
+	        this.paramValues = paramValues;
+	    }
+	  
+		@Override
+		public Stream<BytesHandle> getParamValues() {
+			return paramValues;
```
ehennum: The getter could then fulfill the contract by constructing a stream on request using Stream.of() similar to MultipleAtomicCallField.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -450,37 +464,116 @@ public SingleAtomicCallField(String paramName, String paramValue) {
...
+    	return new BufferedMultipleNodeCallField(super.paramName, Stream.of(bytesHandle));
+    }
+    
+  }
+  static public class BufferedMultipleAtomicCallField extends CallField implements MultipleAtomicCallField {
+	  
+	private String[] paramValues;
+	public BufferedMultipleAtomicCallField(String paramName, Stream<String> paramValues) {
+		super(paramName);
+        this.paramValues = (String[]) paramValues.toArray();
```
ehennum: My understanding is that the preferred way to construct the array is using the lambda, as in:

    `paramValues.toArray(size -> new String[size])`

That way, the array is constructed once with the correct type and the correct number of elements for filling by the stream.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -450,37 +464,116 @@ public SingleAtomicCallField(String paramName, String paramValue) {
...
     }
+    
+    @Override
     public Stream<? extends AbstractWriteHandle> getParamValues() {
       return paramValues;
     }
+    
+    @Override
+    public BufferedMultipleNodeCallField toBuffered() {
+    	if(!(paramValues instanceof BufferableHandle))
```
ehennum: paramValues is a stream containing BufferableHandle items and not a BufferableHandle itself.  So, the cast will fail.

Instead, it will be necessary to convert the items in the stream individually.

One way to do that would be with a mapping lambda, as in:

```
    Stream<BytesHandle> bytesHandleStream = paramValues.map(handle ->{
        ...  if the handle is not a BufferableHandle, throw an exception, otherwise, construct the BytesHandle ...
    });
```

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -455,13 +460,24 @@ CallArgsImpl addDefaultArgs(CallArgsImpl callArgsImpl) {
 		if(defaultArgs == null || defaultArgs.size()==0)
 			return callArgsImpl;
 		CallerImpl callerImpl = (com.marklogic.client.dataservices.impl.CallManagerImpl.CallerImpl) caller;
-		List<CallField> newCallFields = new ArrayList<CallField>();
+		Set<CallField> newCallFields = new HashSet<CallField>();
```
ehennum: Instead of converting newCallFields to a List, does it make more sense to create it as a list originally?

Would that make line 479 unnecessary?

marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java
```diff
@@ -541,37 +553,34 @@ public BufferedMultipleAtomicCallField(String paramName, Stream<String> paramVal
...
+		@Override
+		public BytesHandle getParamValue() {
+			return paramValue;
 		}
 
-		@Override
-		public AbstractWriteHandle getParamValue() {
-			// TODO Auto-generated method stub
-			return null;
+		public Stream<AbstractWriteHandle> getParamValues() {
```
ehennum: Given that SingleAtomicCallField doesn't have a stream getter, what is the rationale for SingleNodeCalleField to have a stream getter?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -423,14 +424,18 @@ public Boolean call() {
 	public CallBatcherImpl<W, E> withdefaultArgs(CallArgs args) {
 		if(args == null)
 			throw new IllegalArgumentException("CallArgs cannot be null.");
+		
 		CallArgsImpl callArgsImpl = (CallArgsImpl) args;
+		
 		if(callArgsImpl == null || ! (callArgsImpl instanceof CallArgsImpl))
 			throw new IllegalArgumentException("Illegal Argument.");
```
ehennum: If callArgsImpl is null, should we return this for consistency with line 434? 

I believe the instanceof test must precede the cast.  Otherwise, the cast will fail before the instanceof test is executed.

PR #1081, opened 2019-04-10
unit tests for default parameters

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-60, https://project.marklogic.com/jira/browse/JAVA-61
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
@@ -102,7 +102,6 @@ public interface CallBatcher<W,E extends CallManager.CallEvent> extends Batcher 
      * for a single parameter.
      * @param args  one or more parameters for the caller
      * @return  the CallBatcher for chaining with other configuration methods
-     */
     CallBatcher<W,E> withDefaultArgs(CallManager.CallArgs args);
 
     /**
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
@@ -103,7 +103,7 @@ public interface CallBatcher<W,E extends CallManager.CallEvent> extends Batcher 
      * @param args  one or more parameters for the caller
      * @return  the CallBatcher for chaining with other configuration methods
      */
-    CallBatcher<W,E> withdefaultArgs(CallManager.CallArgs args);
+    CallBatcher<W,E> withDefaultArgs(CallManager.CallArgs args);
 
     /**
      * Gets the listeners that process results for successful calls.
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherDefaultArgsTest.java
```diff
@@ -98,8 +98,8 @@ public void singleAtomicCallFieldTest() {
                     fail("atomic batcher failed");
                     });
 
-        batcher.startJob​();
-        batcher.add(caller.args().param("param1", 1.1));
+        batcher.getDataMovementManager().startJob(batcher);
```
ehennum: The start() in the previous version of line 101 is supported (and probably better than the revision because simpler).

Also, just so you know, adding arguments implicitly starts a job.

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/CallBatcherDefaultArgsTest.java
```diff
@@ -321,10 +321,6 @@ public void singleAtomicDifferentParamsTest() {
                     fail("atomic batcher failed");
                 });
```
ehennum: If the test doesn't provide any data, what is being tested?  Do I misunderstand?

anu3990: The "twoDiffAtomic" endpoint in singleAtomicDifferentParamsTest is configured for taking two parameters, first a double (multiple) which is not nullable and a float value(not multiple) which can be nullable. So this test case is asserting that no exception is thrown when a null float value is passed as second parameter. 
The next test case singleAtomicDifferentParamsTest2 throws an exception on passing a null float value because there the second parameter is configured to be not nullable.

anu3990:
Also adding javadoc for withDefaultArgs.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
```diff
@@ -33,6 +33,12 @@
     CallBatcher<W,E> withJobId(String jobId);
     CallBatcher<W,E> withJobName(String jobName);
     CallBatcher<W,E> withThreadCount(int threadCount);
+    
+    /**
+     * Adds the CallArgs to the current instance of CallBatcher.
```
ehennum: The default argument for a parameter is used only when the parameter isn't otherwise specified.

Would it be useful to explain this to the API user?  If one were using this API, what would be helpful to understand?

anu3990: Thanks Erik.! 
Adding the explanation for default args.

PR #1083, opened 2019-04-15
Implementing an argument generator callback.

anu3990:
https://project.marklogic.com/jira/browse/JAVA-52
* Are you modifying the correct branch? Yes - develop
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
@@ -16,12 +16,15 @@
 package com.marklogic.client.dataservices;
 
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 import com.marklogic.client.datamovement.Batcher;
 import com.marklogic.client.datamovement.DataMovementManager;
 import com.marklogic.client.datamovement.ForestConfiguration;
 import com.marklogic.client.datamovement.JobTicket;
+import com.marklogic.client.dataservices.CallManager.CallArgs;
+import com.marklogic.client.dataservices.CallManager.CallEvent;
 
 /**
  * A CallBatcher executes multiple concurrent calls to a Data Service endpoint
@@ -276,5 +279,11 @@ public interface CallBatcher<W,E extends CallManager.CallEvent> extends Batcher 
          * @return  a CallBatcher that accepts the input and makes calls with the caller
          */
         CallBatcher<CallManager.CallArgs,E> forArgs();
+        
+        CallBatcher<CallArgs,E> forArgsGenerator(CallArgsGenerator<E> generator);
+    }
+    
+    @FunctionalInterface
+    public interface CallArgsGenerator<E extends CallEvent> extends Function<E, CallArgs> {
     }
 }
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
@@ -280,7 +280,7 @@ public interface CallBatcher<W,E extends CallManager.CallEvent> extends Batcher 
          */
         CallBatcher<CallManager.CallArgs,E> forArgs();
         
-        CallBatcher<CallArgs,E> forArgsGenerator(CallArgsGenerator<E> generator);
+        CallBatcher<Void,E> forArgsGenerator(CallArgsGenerator<E> generator);
     }
     
     @FunctionalInterface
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/CallBatcher.java
```diff
@@ -276,5 +279,11 @@
          * @return  a CallBatcher that accepts the input and makes calls with the caller
          */
         CallBatcher<CallManager.CallArgs,E> forArgs();
+        
+        CallBatcher<CallArgs,E> forArgsGenerator(CallArgsGenerator<E> generator);
```
ehennum: The technical story claims that forArgsGenerator() should return

`CallBatcher<Void,E>`

on the grounds that add() should not be supported as a way of providing input.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -287,6 +293,8 @@ public CallBatcher add(W input) {
 
         CallManagerImpl.CallArgsImpl args = null;
         if (input == null) {
+            if(inputType == Void.class)
+                throw new IllegalArgumentException("Input type is void");
```
ehennum: A more useful error message might be something like:  "cannot call add() when supplying arguments with generator"

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -457,7 +465,16 @@ private void initialize() {
             logger.warn("threadCount should be 1 or greater -- setting threadCount to number of hosts: " + threadCount);
         }
         if (initialized.getAndSet(true)) return;
-// TODO
+
```
ehennum: The thread pool should be initialized before the for loop starts submitting tasks to execute on the thread pool.

In other words, consider moving the test and for loop after setting the started flag.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -457,7 +465,16 @@ private void initialize() {
             logger.warn("threadCount should be 1 or greater -- setting threadCount to number of hosts: " + threadCount);
         }
         if (initialized.getAndSet(true)) return;
-// TODO
+        
+        if(inputType == Void.class) {
```
ehennum: In looking at the code, I'm wondering belatedly whether the operation might be easier to understand if the test was on

`callArgsGenerator != null`

instead of on 

    `inputType == Void.class` 

both here and also on line 708 (but keeping the Void test on line 296).

What do you think?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -553,6 +570,11 @@ private void retry(CallManager.CallEvent event, boolean callFailListeners) {
         public CallBatcherImpl<CallManager.CallArgs, E> forArgs() {
             return new CallBatcherImpl(client, caller, CallManager.CallArgs.class).finishConstruction();
         }
+
+        @Override
+        public CallBatcher<CallArgs, E> forArgsGenerator(CallArgsGenerator<E> generator) {
+            return new CallBatcherImpl(client, caller, java.lang.Void.class, generator);
```
ehennum: Given that Void is imported in the file, is there any reason to use the package name prefix?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -683,6 +705,12 @@ public Boolean call() {
 
             initEvent(output, callTime);
             batcher.sendSuccessToListeners(output);
+            if(batcher.inputType == Void.class) {
+                CallArgs newInput = batcher.callArgsGenerator.apply(output);
```
ehennum: We need to consider how to respond to errors in the user code in the call generator and, for that matter, the listeners.

Please add a TODO to check how WriteBatcher and QueryBatcher tolerate listener errors.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
anu3990: Yes. we can do that.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/CallBatcherImpl.java
```diff
@@ -466,18 +466,18 @@ private void initialize() {
...
-        if(inputType == Void.class) {
+        threadPool = new CallingThreadPoolExecutor(this, getThreadCount());
+        jobStartTime = Calendar.getInstance();
+        started.set(true);
+        
+        if(callArgsGenerator != null) {
             for (int i=0;i<getThreadCount(); i++) {
                     CallArgs newInput = callArgsGenerator.apply(null);
                     if(newInput != null) {
                         submitCall((CallArgsImpl) newInput);
```
ehennum: Before submitCall() here and on line 711, should there be a test that throws an "unsupported implementation of call arguments" exception if CallArgs is not an instance of CallArgsImpl?

PR #1088, opened 2019-04-23
send call generator errors to failure listeners

ehennum:


ehennum:
This pull request proposes wrapping errors from user-supplied call generators in an special exception for sending to failure listeners so listeners can distinguish call generator errors from failed calls.

I'm not entirely convinced that this is the right solution.  After all, the developer could just as easily wrap all of the code inside a call generator in a try / catch.  Also, currently, failure listeners always respond to errors outside the developer's control.

Let's discuss tomorrow.

ehennum:
On fully convinced this is not necessary or appropriate.

PR #1108, opened 2019-05-31
Adding the conversion methods

anu3990:
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/BaseProxy.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/BaseProxy.java
@@ -104,7 +104,7 @@ public class BaseProxy {
       static final public String         fromBoolean(Boolean value)                    { return ValueConverter.BooleanToString(value);  }
       static final public Stream<String> fromBoolean(Stream<? extends Boolean> values) { return ValueConverter.BooleanToString(values); }
       static final public String[] fromBoolean(Boolean[] values) { 
-          return (Stream.of(values).map(value->ValueConverter.BooleanToString(value)).toArray(String[]::new)); 
+          return ValueConverter.convert(values, ValueConverter::BooleanToString); 
       }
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/BaseProxy.java
```diff
@@ -93,8 +104,7 @@ public BaseProxy(DatabaseClient db, String endpointDir) {
       static final public String         fromBoolean(Boolean value)                    { return ValueConverter.BooleanToString(value);  }
       static final public Stream<String> fromBoolean(Stream<? extends Boolean> values) { return ValueConverter.BooleanToString(values); }
       static final public String[] fromBoolean(Boolean[] values) { 
-          String[] stringValues = Stream.of(values).map(value->ValueConverter.BooleanToString(value)).toArray(String[]::new);
-          return NodeConverter.arrayWithFormat(stringValues , Format.getFromMimetype(NAME)); 
+          return (Stream.of(values).map(value->ValueConverter.BooleanToString(value)).toArray(String[]::new));
```
ehennum: To get rid of the streams, I would propose adding this convenience to ValueConverter:

```
  static public <I> String[] convert(I[] in, Function<I, String> converter) {
    if (in == null) return null;
    String[] out = new String[in.length];
    for (int i=0; i < in.length; i++) {
      out[i] = converter.apply(in[i]);
    }
    return out;
  }
```

and then adding the following array converters in ValueConverter:

```
  static public String[] BooleanToString(Boolean[] values) {
    return convert(values, ValueConverter::BooleanToString);
  }

```
and then calling them from BaseProxy similar to

`    return ValueConvert.BooleanToString(values);`

Does that make sense?

anu3990: yes. That's right.

Issue #1095, opened 2019-05-07
Calling stop() in a QueryBatcher backed by iterator shuts it down before the running listeners complete


srinathgit:
So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt
4.2.0
### Version of MarkLogic Server
9.0-9
See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version
Java 8
Run `java -version`

### OS and version
MAC OS X 10.13.6
For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

### Input: Some code to illustrate the problem, preferably in a state that can be independently reproduced on our end

https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```
 queryBatcher = dataMovementManager.newQueryBatcher(uris.iterator())
            .withBatchSize(batchSize)
            .withThreadCount(threadCount)
            .withJobId(runStepResponse.getJobId())
            .onUrisReady((QueryBatch batch) -> {
                try {
                    //FlowResource is REST extension that does harmonization step
                    FlowResource flowResource;
                    Map<String,Object> optsMap = new HashMap<>(options);
                    if (databaseClientMap.containsKey(batch.getClient())) {
                        flowResource = databaseClientMap.get(batch.getClient());
                    } else {
                        flowResource = new FlowResource(batch.getClient(), destinationDatabase, flow);
                        databaseClientMap.put(batch.getClient(), flowResource);
                    }
                    optsMap.put("uris", batch.getItems());

                    ResponseHolder response = flowResource.run(runStepResponse.getJobId(), step, optsMap);
                    failedEvents.addAndGet(response.errorCount);
                    successfulEvents.addAndGet(response.totalCount - response.errorCount);
                    
                } catch (Exception e) {
                    if (errorMessages.size() < MAX_ERROR_MESSAGES) {
                        errorMessages.add(e.toString());
                    }
                }
            })
            .onQueryFailure((QueryBatchException failure) -> {
                failedBatches.addAndGet(1);
                failedEvents.addAndGet(batchSize);
            });

        if(! isStopped.get()) {
            JobTicket jobTicket = dataMovementManager.startJob(queryBatcher);
[...]

PR #1112, opened 2019-06-07
Limiting the maximum number of uris that can be collected.

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1095
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
@@ -89,7 +89,8 @@ public class QueryBatcherImpl extends BatcherImpl implements QueryBatcher {
   private JobTicket jobTicket;
   private Calendar jobStartTime;
   private Calendar jobEndTime;
-  private long maxUris;
+  private long maxUris = Long.MAX_VALUE;
+  private AtomicBoolean jobDone = new AtomicBoolean(false);
 
   public QueryBatcherImpl(QueryDefinition query, DataMovementManager moveMgr, ForestConfiguration forestConfig) {
     super(moveMgr);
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
```diff
@@ -89,6 +89,7 @@
   private JobTicket jobTicket;
   private Calendar jobStartTime;
   private Calendar jobEndTime;
+  private long maxUris;
```
ehennum: Should maxUris be initialized to Long.MAX_VALUE so that, by default, it doesn't constrain collection of results?

Alternatively, should maxUris have a Long data type and be initialized to null so that tests can ignore maxUris when it hasn't been set?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
```diff
@@ -127,8 +128,11 @@ public QueryBatcherImpl(Iterator<String> iterator, DataMovementManager moveMgr,
   @Override
   public QueryBatcherImpl onUrisReady(QueryBatchListener listener) {
     if ( listener == null ) throw new IllegalArgumentException("listener must not be null");
-    urisReadyListeners.add(listener);
-    return this;
+    if(resultsSoFar.longValue()< getMaxUris()) {
+        urisReadyListeners.add(listener);
+        return this;
+    }
+    return null;
```
ehennum: I believe the expectation is that listeners will be configured before the job starts.  That is, the resultsSoFar would almost always be zero.  If so, it wouldn't be beneficial to check against the maximum in this method.

Also, fluent method should never return null because chained calls such as withBatchSize() in this example 

    .onUrisReady(...)
    .withBatchSize(...)
 
 would be guaranteed to produce a null pointer exception.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
```diff
@@ -610,7 +608,7 @@ private synchronized void startQuerying() {
     public void run() {
       // don't proceed if this forest is marked as done (because we already got the last batch)
       AtomicBoolean isDone = forestIsDone.get(forest);
-      if ( isDone.get() == true ) {
+      if ( isDone.get() == true || jobDone.get()) {
```
ehennum: I notice the log message isn't appropriate for the jobDone case.  Maybe the test for jobDone should be a separate case instead of an OR?

anu3990: Yes. That's right.

Issue #636, opened 2017-01-20
Test connection


jmakeig:
As a developer I want to debug connection issues. I don't want to have to do any real work (i.e. run a dummy query or test the existence of a document) because that's extra work and potentially introduces other errors. I just want to see if my auth*n and network are configured correctly. 

As an ETL tool maker, I want to provide a "Test Connection" button in my UI.

We should be able to return more than just a Boolean pass/fail. Something like server and client version would be useful. (What else?)

The Client REST API has the undocumented (?) `/v1/ping` endpoint that might be related.

Related to marklogic/node-client-api#346.

jmakeig:
```java
package com.marklogic.datamovement.examples;

import com.marklogic.client.DatabaseClient;
import com.marklogic.client.DatabaseClientFactory;
import com.marklogic.client.DatabaseClientFactory.SecurityContext;
import com.marklogic.client.document.DocumentDescriptor;
import com.marklogic.client.document.DocumentManager;

public class TestDefaultLocalhostConnection {

  public static void main(String[] args) {
    final SecurityContext auth = new DatabaseClientFactory.DigestAuthContext(
        "user", "********");
    final DatabaseClient marklogic = DatabaseClientFactory.newClient("localhost",
        8000, auth);
    System.out.println(marklogic.getHost());
    final DocumentDescriptor doc = marklogic.newDocumentManager().exists("¯\_(ツ)_/¯");
    System.out.println(doc); // null
  }

}
```

sharpau:
I was looking for this very thing in the Java API. Something like exposing the /v1/ping endpoint is exactly what I want.

georgeajit:
Tests run fine.

rjrudin:
Can this be closed since the "checkConnection" method now exists in DatabaseClient?

ehennum:
Most definitely.

PR #1123, opened 2019-08-23
Adding "Test Connection" feature.

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/636
* Are you modifying the correct branch? yes - develop
* Version of MarkLogic Java Client API 5.0
* Version of MarkLogic Server 10.0
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
@@ -27,6 +27,7 @@ import com.marklogic.client.document.TextDocumentManager;
 import com.marklogic.client.document.XMLDocumentManager;
 import com.marklogic.client.eval.ServerEvaluationCall;
 import com.marklogic.client.extensions.ResourceManager;
+import com.marklogic.client.impl.OkHttpServices.ConnectionResult;
 import com.marklogic.client.query.QueryManager;
 import com.marklogic.client.row.RowManager;
 import com.marklogic.client.alerting.RuleManager;
@@ -225,6 +226,13 @@ public interface DatabaseClient {
    * @return the connection type
    */
   ConnectionType getConnectionType();
+  
+  /**
+   * Checks if the connection is valid.
+   * @return {@link ConnectionResult}  with a connected property of true or false. 
+   *  In the false case it contains the errorMessage property identifying the failure.
+   */
+  ConnectionResult checkConnection();
 
   String getHost();
 
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
@@ -27,7 +27,6 @@ import com.marklogic.client.document.TextDocumentManager;
 import com.marklogic.client.document.XMLDocumentManager;
 import com.marklogic.client.eval.ServerEvaluationCall;
 import com.marklogic.client.extensions.ResourceManager;
-import com.marklogic.client.impl.OkHttpServices.ConnectionResult;
 import com.marklogic.client.query.QueryManager;
 import com.marklogic.client.row.RowManager;
 import com.marklogic.client.alerting.RuleManager;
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -6270,4 +6292,28 @@ protected void set() {
         is = true;
     }
   }
+
```
ehennum: Given that ConnectionResult most be visible to the application, should ConnectionResult be a nested static public interface of DatabaseClient with getters for the properties?  

Then, the existing ConnectionResult could be renamed as ConnectionResultImpl, implement the ConnectionResult interface, annotate the getters with @override, and change the visibility on setters to private.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -6270,4 +6292,28 @@ protected void set() {
         is = true;
     }
   }
+  
+  public class ConnectionResult {
```
ehennum: Would the code be easier to read and understand with correct indentation of the private fields and the first method?

marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
```diff
@@ -225,6 +226,13 @@
    * @return the connection type
    */
   ConnectionType getConnectionType();
+  
+  /**
+   * Checks if the connection is valid.
+   * @return {@link ConnectionResult}  with a connected property of true or false.
```
ehennum: Could the @link be removed if the ConnectionResult were a nested interface?  (See comment below)

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -6270,4 +6292,28 @@ protected void set() {
...
+	  private boolean connected = false;
+	  private int statusCode;
```
ehennum: Should statusCode be an Integer instead of an int so it can be null?

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -1244,25 +1242,29 @@ public DocumentDescriptor head(RequestLogger reqlog, String uri,
...
+  @Override
+  public ConnectionResult checkConnection() {
+	  Request.Builder request = new Request.Builder()
+		        .url(this.baseUri);
+	  Response response = headImplExec(null, this.baseUri.uri().toString(), null, request);
+	  ConnectionResult connectionResult = new ConnectionResult();
+	  int statusCode = response.code();
+	  if(statusCode < 300) {
+		  connectionResult.setConnected(true);
+	      connectionResult.setStatusCode(statusCode);
```
ehennum: In the Node.js case, we only set the status code in the false case.  Can we take the same approach here by representing the status code in the field and methods of the ConnectionResult as an Integer?

marklogic-client-api/src/test/java/com/marklogic/client/test/DatabaseClientTest.java
```diff
@@ -116,4 +120,30 @@ public void testGetClientImplementationObject() {
...
+  @Test
+  public void testCheckConnectionWithValidUser() {
+		
+	DatabaseClient marklogic = DatabaseClientFactory.newClient("localhost",
+		        8000, "unittest-nodeapi", new DatabaseClientFactory.DigestAuthContext(
+				        "admin", "admin"));
+  
+    ConnectionResult connResult = marklogic.checkConnection();
+    assertTrue(connResult.isConnected());
+    assertTrue(connResult.getStatusCode() < 300);
```
ehennum: Can we remove the assertion on line 133 if the status code can be null in the successful case?

marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClient.java
anu3990: Changed the link to point the interface ConnectionResult.

PR #1127, opened 2019-09-09
Adding changes for QueryBatcher for max uris collection.

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1095
* Are you modifying the correct branch? yes - develop
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/QueryBatcher.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/QueryBatcher.java
@@ -369,9 +369,9 @@ public interface QueryBatcher extends Batcher {
   /**
    * Sets the limit for the maximum number of uris that can be collected.
    * 
-   * @param results is the value of the limit.
+   * @param maxBatches is the value of the limit.
    */
-  void setMaxUris(long results);
+  void setMaxBatches(long maxBatches);
   
   /**
    * Resets the max number of uris limit so the new job can set it to its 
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/QueryBatcher.java
```diff
@@ -380,9 +380,16 @@
   void setMaxUris();
```
ehennum: Did we decide to remove all setters and getter for max uris from the public interface?

Should we have a zero-argument setMaxBatches() that caps the query at the current batch?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
```diff
@@ -91,6 +91,8 @@
   private Calendar jobEndTime;
   private long maxUris = Long.MAX_VALUE;
   private AtomicBoolean jobDone = new AtomicBoolean(false);
+  private long maxBatches = Long.MAX_VALUE;
+  private boolean isMaxBatchesSet = false;
```
ehennum: Instead of a boolean field, could we test maxBatches < Long.MAX_Value as the guard on setMaxUris()?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/QueryBatcherImpl.java
```diff
@@ -1040,7 +1047,12 @@ public void setMaxUris() {
 }
 
 @Override
-public long getMaxUris() {
-    return this.maxUris;
+public void setMaxUris(long maxBatches) {
```
ehennum: Could setMaxUris() be private?

PR #1139, opened 2019-10-02
InputEndpointImpl method implementation and javadoc.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-85 and https://project.marklogic.com/jira/browse/JAVA-87
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
@@ -16,16 +16,25 @@
 package com.marklogic.client.dataservices.impl;
 
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.stream.Stream;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.MarkLogicInternalException;
+import com.marklogic.client.SessionState;
 import com.marklogic.client.dataservices.InputEndpoint;
 import com.marklogic.client.dataservices.impl.IOEndpointImpl;
+import com.marklogic.client.dataservices.impl.IOEndpointImpl.BulkIOEndpointCallerImpl.WorkPhase;
 import com.marklogic.client.io.marker.JSONWriteHandle;
 
 public class InputEndpointImpl extends IOEndpointImpl implements InputEndpoint {
+	private static Logger logger = LoggerFactory.getLogger(InputEndpointImpl.class);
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -83,12 +92,53 @@ private int getBatchSize() {
 
 		@Override
 		public void accept(InputStream input) {
-			// TODO Auto-generated method stub
+			
+			InputStream output = null;
```
ehennum: Would it be possible to declare the variable at the point of assignment on line 130?

If so, that might make the code easier to understand for someone else who maintains it.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -83,12 +92,53 @@ private int getBatchSize() {
...
+			try {
+				queue.put(input);
+			} catch (InterruptedException e) {
+				e.printStackTrace();
```
ehennum: If the bulk caller can't accept the input, is that a condition that the caller would need to know about?

Would it make sense to rethrow the exception as a wrapped runtime exception?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -83,12 +92,53 @@ private int getBatchSize() {
...
+					break;
+				case RUNNING:
+					break;
+				case INTERRUPTING:
+					throw new MarkLogicInternalException(
+							"cannot accept more input as current phase is  " + getPhase().name());
+				case INTERRUPTED:
+					throw new MarkLogicInternalException(
+							"cannot accept more input as current phase is  " + getPhase().name());
+				case COMPLETED:
```
ehennum: Do these cases need separate throw actions, given that cases fall through in Java?

Also, is it accurate to report the problem as an internal exception?  Would a Java IllegalStateException be more accurate?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -83,12 +92,53 @@ private int getBatchSize() {
...
+			
+			logger.trace("input endpoint running endpoint={} count={} state={}",
+                    getEndpointPath(), getCallCount(), getEndpointState());
+			output = getEndpoint().getCaller().call(
+					getEndpoint().getClient(), getEndpointState(), session, getWorkUnit(), inputStreamList.stream()
+					);
+			
+			if (allowsEndpointState()) {
+                setEndpointState(output);
+            }
```
ehennum: If the indentation were more regular, the code might be easier for someone else to maintain.

anu3990:
Added the suggestions and adjusted the indentation.

PR #1143, opened 2019-10-08
Unit Test for bulk InputEndpoint implementation.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-86
* Are you modifying the correct branch? develop-72
* Java version 11.0.4
* OS and version

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkInputCallerTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkInputCallerTest.java
@@ -1,0 +1,121 @@
+package com.marklogic.client.test.dataservices;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.dataservices.ExecEndpoint;
+import com.marklogic.client.dataservices.InputEndpoint;
+import com.marklogic.client.document.JSONDocumentManager;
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkInputCallerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    static ObjectNode apiObj;
+    static String apiName = "bulkInputCallerImpl.api";
+    static String scriptPath;
+    static String apiPath;
+    static int workMax = 5;
+    static int startValue = 1;
+    static JSONDocumentManager docMgr;
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
```
ehennum: Would it be more robust to initialize expectedException to ExpectedException.none() within bulkInputEndpointTest()?

If someone added a test that expects an exception before bulkInputEndpointTest(), would that cause bulkInputEndpointTest() to fail?

anu3990: Hi Erik. I guess it should be okay. I tried adding a unit test to throw null pointer before bulkInputEndpointTest() and it ran successfully. Also, if we move the initialization of expectedException to bulkInputEndpointTest() we will need to initialize it again within bulkInputEndpointInterruptTest().

PR #1147, opened 2019-10-10
Output Endpoint Implementation.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-90 and https://project.marklogic.com/jira/browse/JAVA-91
* Are you modifying the correct branch? develop-72
* Java version 11.0.4
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
@@ -31,7 +31,6 @@ public class OutputEndpointImpl extends IOEndpointImpl implements OutputEndpoint
     private static Logger logger = LoggerFactory.getLogger(OutputEndpointImpl.class);
     private OutputCallerImpl caller;
 
-
     public OutputEndpointImpl(DatabaseClient client, JSONWriteHandle apiDecl) {
         this(client, new OutputCallerImpl(apiDecl));
     }
@@ -61,7 +60,16 @@ public class OutputEndpointImpl extends IOEndpointImpl implements OutputEndpoint
 
         @Override
         public void awaitCompletion() {
+            if(outputConsumer == null)
+                return;
+
             logger.trace("output endpoint running endpoint={} work={}", getEndpointPath(), getWorkUnit());
+
+            if(getPhase() != WorkPhase.INITIALIZING) {
+                throw new IllegalStateException(
+                        "Cannot process output since current phase is  " + getPhase().name());
+            }
+
             setPhase(WorkPhase.RUNNING);
             SessionState session = allowsSession() ? getEndpoint().getCaller().newSessionState() : null;
 
@@ -79,14 +87,17 @@ public class OutputEndpointImpl extends IOEndpointImpl implements OutputEndpoint
                     throw new RuntimeException("error while calling "+getEndpoint().getEndpointPath(), throwable);
                 }
 
+                InputStream[] result = null;
+                if(output != null) {
+                    result = output.toArray(size -> new InputStream[size]);
+                }
                 if (allowsEndpointState()) {
-                    if(output!= null && output.findFirst()!= null && output.findFirst().isPresent()) {
-                        InputStream[] result = output.toArray(size -> new InputStream[size]);
-                        setEndpointState(result[0]);
-                        for(int i=1; i<result.length; i++) {
-                            forEachOutput((Consumer<InputStream>) result[i]);
+                        if(result !=null) {
+                            setEndpointState(result[0]);
                         }
-                    }
+                }
+                for(int i=1; i<result.length; i++) {
+                    outputConsumer.accept(result[i]);
                 }
 
                 switch(getPhase()) {
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
@@ -16,6 +16,8 @@
 package com.marklogic.client.dataservices.impl;
 
 import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.MarkLogicInternalException;
+import com.marklogic.client.SessionState;
 import com.marklogic.client.dataservices.OutputEndpoint;
 import com.marklogic.client.io.marker.JSONWriteHandle;
 import org.slf4j.Logger;
@@ -23,6 +25,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.InputStream;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 public class OutputEndpointImpl extends IOEndpointImpl implements OutputEndpoint {
     private static Logger logger = LoggerFactory.getLogger(OutputEndpointImpl.class);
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -59,7 +62,59 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
...
+                    output = getEndpoint().getCaller().call(
+                            getEndpoint().getClient(), getEndpointState(), session, getWorkUnit()
+                    );
+                    incrementCallCount();
+                } catch(Throwable throwable) {
+                    throw new RuntimeException("error while calling "+getEndpoint().getEndpointPath(), throwable);
+                }
+
+                if (allowsEndpointState()) {
+                    if(output!= null && output.findFirst()!= null && output.findFirst().isPresent()) {
```
ehennum: findFirst() is a terminal operation that completely consumes the stream.

Would it work to perform the tests in steps?

That is, if the output is not null, construct the array.  Then, if the array is not empty, process it?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -59,7 +62,59 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
...
+
+                    output = getEndpoint().getCaller().call(
+                            getEndpoint().getClient(), getEndpointState(), session, getWorkUnit()
+                    );
+                    incrementCallCount();
+                } catch(Throwable throwable) {
+                    throw new RuntimeException("error while calling "+getEndpoint().getEndpointPath(), throwable);
+                }
+
+                if (allowsEndpointState()) {
```
ehennum: Should the client also process the output if there is no endpoint state?

Would it work to move the construction of the result array before this if branch and the for loop after the if branch.

The if branch would also need to increment i from zero to 1.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -59,7 +62,59 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
...
+                    if(output!= null && output.findFirst()!= null && output.findFirst().isPresent()) {
+                        InputStream[] result = output.toArray(size -> new InputStream[size]);
+                        setEndpointState(result[0]);
+                        for(int i=1; i<result.length; i++) {
+                            forEachOutput((Consumer<InputStream>) result[i]);
```
ehennum: forEachOutput() is a setter for a Consumer function.  The loop should call the Consumer function's accept() method.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -59,7 +62,59 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
         @Override
         public void awaitCompletion() {
```
ehennum: Would it be appropriate to add a guard to make sure the output consumer is not null?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -59,7 +62,59 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
...
             logger.trace("output endpoint running endpoint={} work={}", getEndpointPath(), getWorkUnit());
+            setPhase(WorkPhase.RUNNING);
```
ehennum: Should we add a test that the current WorkPhase is INITIALIZING and throw an illegal state exception if it isn't?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -61,7 +60,16 @@ private BulkOutputCallerImpl(OutputEndpointImpl endpoint) {
 
         @Override
         public void awaitCompletion() {
+            if(outputConsumer == null)
+                return;
```
ehennum: Should it throw an exception instead?  If the application hasn't supplied a consumer, is it meaningful to await completion?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -79,14 +87,17 @@ public void awaitCompletion() {
                     throw new RuntimeException("error while calling "+getEndpoint().getEndpointPath(), throwable);
                 }
 
+                InputStream[] result = null;
```
ehennum: Should the result variable and all of the result processing move inside the branch that tests output is not null?  If there's no output, by definition, there's no result processing.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -79,14 +87,17 @@ public void awaitCompletion() {
...
+                if(output != null) {
+                    result = output.toArray(size -> new InputStream[size]);
+                }
                 if (allowsEndpointState()) {
-                    if(output!= null && output.findFirst()!= null && output.findFirst().isPresent()) {
-                        InputStream[] result = output.toArray(size -> new InputStream[size]);
-                        setEndpointState(result[0]);
-                        for(int i=1; i<result.length; i++) {
-                            forEachOutput((Consumer<InputStream>) result[i]);
+                        if(result !=null) {
```
ehennum: Should there be a test that the result.length is greater than 0?  Instead of nested if branches, would the code be easier to read as an && condition on the test of allowsEndpointState()?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -79,14 +87,17 @@ public void awaitCompletion() {
...
+                            setEndpointState(result[0]);
                         }
-                    }
+                }
+                for(int i=1; i<result.length; i++) {
```
ehennum: Should i be initialized to 0 if the allowsEndpointState() is false?  In that case, the first result would be output instead of the endpoint state.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -87,17 +87,15 @@ public void awaitCompletion() {
...
 
-                InputStream[] result = null;
                 if(output != null) {
-                    result = output.toArray(size -> new InputStream[size]);
-                }
-                if (allowsEndpointState()) {
-                        if(result !=null) {
+                    InputStream[] result = output.toArray(size -> new InputStream[size]);
+
+                    if (allowsEndpointState() && result != null && result.length>0) {
```
ehennum: As I understand Stream, if output is not null, the output.toArray() is guaranteed to be not null.

More generally, either the if test doesn't need to test for result != null or the for loop should have the same test.

PR #1148, opened 2019-10-16
Adding logic to awaitcompletion to process the last batch.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-85
* Version of MarkLogic Java Client API (see Readme.txt) 5.0.1
* Java version (`java -version`) 11.0.4
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
@@ -120,20 +120,19 @@ public class InputEndpointImpl extends IOEndpointImpl implements InputEndpoint {
 				throw new MarkLogicInternalException(
 						"unexpected state for " + getEndpointPath() + " during loop: " + getPhase().name());
 			}
-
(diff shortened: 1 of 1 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -120,24 +120,31 @@ public void accept(InputStream input) {
...
+		}
+
+		@Override
+		public void awaitCompletion() {
+			logger.trace("input endpoint running endpoint={} work={}", getEndpointPath(), getWorkUnit());
+			if(queue != null && !queue.isEmpty())
+				processInput(queue);
+		}
+
+		private void processInput(LinkedBlockingQueue<InputStream> queue) {
```
ehennum: Is there a rationale for passing the queue as a parameter instead of accessing it locally?

By the same reasoning, is there a rationale for logging in the caller instead of locally?

If not, maybe it would be simpler and easier to maintain to access the queue and to log locally within processInput().

anu3990: Thanks Erik.!
Added the changes.

PR #1156, opened 2019-11-08
Adding unit test for bulk output caller.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-92
* Version of MarkLogic Java Client API - 5.0.1
* Version of MarkLogic Server - 10.0
* Java version- 11
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
@@ -114,7 +114,7 @@ public class OutputEndpointImpl extends IOEndpointImpl implements OutputEndpoint
                                 getEndpointPath(), getCallCount(), getWorkUnit());
                         break calling;
                     case RUNNING:
-                        if (output == null) {
+                        if (output == null || output.length == 0) {
                             setPhase(WorkPhase.COMPLETED);
                             logger.info("output completed endpoint={} count={} work={}",
                                     getEndpointPath(), getCallCount(), getWorkUnit());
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerTest.java
```diff
@@ -0,0 +1,113 @@
...
+        bulkCaller.setWorkUnit(new ByteArrayInputStream(workUnit.getBytes()));
+        bulkCaller.forEachOutput(i-> {
+            try {
+                assertTrue(list.contains(IOTestUtil.mapper.readValue(i, ObjectNode.class).toString()));
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        });
+
+       bulkCaller.awaitCompletion();
```
ehennum: Would it make sense to confirm that forEachOutput() executed the expected number of times?

That is, should the lambda increment a counter and an assertion on the counter follow completion?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -114,7 +114,7 @@ public void awaitCompletion() {
                                 getEndpointPath(), getCallCount(), getWorkUnit());
                         break calling;
                     case RUNNING:
-                        if (output == null) {
+                        if (output == null || output.length == 0) {
```
ehennum: The defensive programming here seems correct.

I'm wondering whether arrayCall() should be returning an empty array or null.  Something to discuss.

PR #1157, opened 2019-11-14
Add BulkOutputCaller.next() for pulling output in Kafka.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-156
* Version of MarkLogic Server - 10
* Java version - 5.0.1
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/OutputEndpoint.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/OutputEndpoint.java
@@ -62,7 +62,8 @@ public interface OutputEndpoint extends IOEndpoint {
          */
         void forEachOutput(Consumer<InputStream> outputConsumer);
         /**
-         * Called by the Kafka dataflow connector to get the next batch of output.
+         * Provides synchronous access to output.
+         * @return the response from the endpoint.
          */
         Stream<InputStream> next();
     }
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/OutputEndpoint.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/OutputEndpoint.java
@@ -61,5 +61,9 @@ public interface OutputEndpoint extends IOEndpoint {
          * @param outputConsumer a function for processing the endpoint output
          */
         void forEachOutput(Consumer<InputStream> outputConsumer);
+        /**
+         * Called by the Kafka dataflow connector to get the next batch of output.
+         */
+        Stream<InputStream> next();
     }
 }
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/OutputEndpoint.java
```diff
@@ -61,5 +61,9 @@ static OutputEndpoint on(DatabaseClient client, JSONWriteHandle apiDecl) {
          * @param outputConsumer a function for processing the endpoint output
          */
         void forEachOutput(Consumer<InputStream> outputConsumer);
+        /**
+         * Called by the Kafka dataflow connector to get the next batch of output.
```
ehennum: Let's not document the new method as exclusive to Kafka -- it's available for any consumer.  

Instead, would it work to clarify that next() provides synchronous access to output and the output consumer provides asynchronous access to output?

Also, could the JavaDoc document the return value?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -274,5 +272,19 @@ public void interrupt() {
...
         }
+
+        void processOutput(InputStream[] output, Consumer<InputStream> outputConsumer) {
+            for (int i=allowsEndpointState()?1:0; i < output.length; i++) {
+                outputConsumer.accept(output[i]);
+            }
+        }
+
+        InputStream[] getOutput(InputStream[] output, Consumer<InputStream> outputConsumer) {
+            processOutput(output, outputConsumer);
```
ehennum: Would it make sense to support either asynchronous consumption of the output via an output consumer or synchronous consumption of the output via next() but not both?

One concern is avoiding the need to buffer the data from each InputStream so that we can resuse the same data for both the synchronous and asynchronous consumption. 

In other words, would it make sense to delete the call on line 283?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -274,5 +272,19 @@ public void interrupt() {
...
+            if(allowsEndpointState()) {
```
ehennum: Should the if condition also make sure that the output.length > 0?

Arrays.copyOfRange(output, 1, 0) might throw an error.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -257,10 +258,7 @@ void processOutputBatch(InputStream[] output, Consumer<InputStream> outputConsum
             if (allowsEndpointState() && output.length > 0) {
                 setEndpointState(output[0]);
             }
```
ehennum: Does the code that caches the endpoint state (lines 258 through 259 above) execute in the next() case or only in the output consumer case?

PR #1162, opened 2019-11-18
Adding unit test for bulkOutputCaller.next().

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-156
* Java version 11
* OS and version

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
@@ -43,7 +43,7 @@ public class BulkOutputCallerNext {
     @Test
     public void bulkOutputCallerNextTest() throws Exception {
         String endpointState = "{\"next\":"+1+"}";
-        String workUnit      = "{\"max\":"+5+"}";
+        String workUnit      = "{\"limit\":"+5+"}";
 
         IOTestUtil.writeDocuments(10,collectionName);
 
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
@@ -15,6 +15,8 @@ import org.junit.rules.ExpectedException;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.stream.Stream;
 
 import static org.junit.Assert.*;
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/test/resources/dataservices/bulkOutputCallerNext.sjs
```diff
@@ -0,0 +1,17 @@
...
+const work = fn.head(xdmp.fromJSON(workUnit));
+const limit = fn.head(xdmp.fromJSON(workUnit));
+
+const res = [];
+const d = fn.subsequence(fn.collection("bulkOutputCallerNext"), state.next, limit.max);
+res.push(state);
+for (const x of d) {
+res.push(x); };
+state.next = state.next + work.max;
+limit.max = limit.max + work.max;
```
ehennum: Is there a reason to increase the limit?

The limit is a constant that specifies the number of results to take after the offset.  

Also, because the work unit has constants and thus isn't returned from the endpoint, modifying the work unit doesn't change the next request.

marklogic-client-api/src/test/resources/dataservices/bulkOutputCallerNext.sjs
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const state  = fn.head(xdmp.fromJSON(endpointState));
+const work = fn.head(xdmp.fromJSON(workUnit));
+const limit = fn.head(xdmp.fromJSON(workUnit));
```
ehennum: Is there a reason to get the work unit twice?  Would it work just as well with work.max instead of limit.max?

Also, would "limit" indicate the purpose of the property better than "max"?

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/IOTestUtil.java
```diff
@@ -87,5 +88,15 @@ public static void load(String apiName, ObjectNode apiObj, String scriptPath, St
     static InputStream asInputStream(String value) {
         return new ByteArrayInputStream(value.getBytes());
     }
+
+    static void writeDocuments(int count, String collection) {
```
ehennum: A good addition to the test utility.

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @AfterClass
+    public static void cleanup() {
+
+        QueryManager queryMgr = IOTestUtil.db.newQueryManager();
+        DeleteQueryDefinition deletedef = queryMgr.newDeleteDefinition();
+        deletedef.setCollections(collectionName);
+        queryMgr.delete(deletedef);
+
+        IOTestUtil.db.release();
```
ehennum: I found that releasing the client had the effect of breaking other tests when they are run at the same time.

Does the test still work if the release() call is deleted?

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/BulkOutputCallerNext.java
```diff
@@ -0,0 +1,79 @@
...
+        output = bulkCaller.next();
+        assertNotNull(output);
+        InputStream[] outputArray1 = output.toArray(size -> new InputStream[size]);
+        assertTrue(outputArray1.length == 5);
+        assertFalse(outputArray.equals(outputArray1));
+        for(int i=0; i<5; i++) {
+            String data = IOTestUtil.mapper.readValue(outputArray[i], ObjectNode.class).toString();
+            String data1 = IOTestUtil.mapper.readValue(outputArray1[i], ObjectNode.class).toString();
+            assertFalse(data.equals(data1));
+        }
```
ehennum: The test would definitely catch some bad results.

Would the test increase our confidence in the implementation, however, if it tested that every document written was retrieved instead of testing that the document at the same result position are different in the two result sets?

For instance, would the following results test as different

    [doc0, doc1, doc2, doc3, doc4] 
    [doc1, doc2, doc3, doc4, doc0] 

when, in fact, the result sets are the same but in a different order?

anu3990: Yes that's right. Changed the test. It now confirms that each index element is not present in the other list.

PR #1164, opened 2019-11-20
changes for DHFPROD-3107.

llinggit:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/DHFPROD-3107
* Version of MarkLogic Java Client API - 5.0.1
* Version of MarkLogic Server - 10
* Java version- 11
* OS and version


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
```diff
@@ -0,0 +1,110 @@
...
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.StringHandle;
+
+import java.io.*;
+import java.nio.charset.Charset;
+import java.util.stream.Stream;
+
+/**
+ * java doc to do
+ */
```
ehennum: A JavaDoc placeholder that slipped through?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
```diff
@@ -0,0 +1,110 @@
...
+    public void setFormat(Format format) {
+        if (format == null) {
+            throw new IllegalArgumentException("some exception");
+        }
+
+        this.format = format;
+    }
+
+    /**
+     *
```
ehennum: The format of the JavaDoc is best with a lead sentence.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Returns the document format set to splitter.
+     * @return the document format set to splitter. The default is Format.JSON.
+     */
+    public Format getFormat() {
+        return this.format;
+    }
+
+    /**
+     * Used to set document format to splitter
```
ehennum: Would it be useful to point that that the format should be set before splitting?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
```diff
@@ -0,0 +1,110 @@
...
+     * @return the number of objects in the stream.
+     */
+    @Override
+    public long getCount() {
+        return count;
+    }
+
+    /**
+     * Takes the input stream and converts it into a stream of StringHandle. The content could be
+     * line-delimited JSON file, line-delimited XML file or gzip-compressed line-delimited JSON file.
```
ehennum: Would it be useful to specifically mention a GZIPInputStream?  Also, wouldn't compression work for XML as well as JSON?

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
```diff
@@ -0,0 +1,100 @@
...
+public class LineSplitterTest {
+    static final private String jsonlFile = "src/test/resources/data" + File.separator + "line-delimited.jsonl";
+    static final private String jsonlGzipFile = "src/test/resources/data" + File.separator + "line-delimited.jsonl.gz";
+    static final private String xmlFile = "src/test/resources/data" + File.separator + "line-delimited.txt";
+    private DatabaseClient client;
+
+    @Before
+    public void setUp() {
+        client = DatabaseClientFactory.newClient("localhost", 8012,
+                new DatabaseClientFactory.DigestAuthContext("rest-admin", "x"));
```
ehennum: If the test uses com.marklogic.client.test.Common.connect() to get a database client, we can maintain the authentication in one place.

Unless there's another reason to use rest-admin, writing and reading should work with the less privileged rest-writer.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertEquals(result[0].getFormat(), Format.XML);
+        for (int i = 0; i < result.length && i < originalResult.length; i++) {
+            assertNotNull(result[i].get());
+            assertEquals(result[i].get(), originalResult[i]);
+        }
+    }
+
+    @After
+    public void closeSetUp() {
+        client.release();
```
ehennum: If the test uses com.marklogic.client.test.Common.connect(), there's no need to release -- other tests can use the same connection.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    public void testSplitter() throws Exception {
+        LineSplitter splitter = new LineSplitter();
+        Stream<StringHandle> contentStream = splitter.split(new FileInputStream(jsonlFile));
+        assertNotNull(contentStream);
+
+        StringHandle[] result = contentStream.toArray(size -> new StringHandle[size]);
+        assertNotNull(result);
+        assertEquals(result.length, Files.lines(Paths.get(jsonlFile)).count());
```
ehennum: The test might run slightly faster if, instead of reading the file twice, the test moved the originalResult up to line 37 as an initialization and used the originalResult.length for this assertion.  Just a thought.

Also, if we have asserted result.length is equal to originalResult.length, we don't need to test both in the loop below.  Just a thought.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertNotNull(result);
+        assertEquals(result.length, Files.lines(Paths.get(xmlFile)).count());
+
+        String[] originalResult = Files.lines(Paths.get(xmlFile))
+                                    .toArray(size -> new String[size]);
+        assertEquals(result[0].getFormat(), Format.XML);
+        for (int i = 0; i < result.length && i < originalResult.length; i++) {
+            assertNotNull(result[i].get());
+            assertEquals(result[i].get(), originalResult[i]);
+        }
```
ehennum: One other thought.

All 3 tests share the same fundamental logic for inspecting the results of splitting.  Would it make sense to create a private function that takes a Stream<StringHandle>, a Format, and the String[] array of file lines and is called from each of the three tests to perform the inspection common to the three tests?

If that's possible, that would make maintenance easier (with the cost of having to break out separate tests if we ever discover a need for different inspection down the road).

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
llinggit: I added below txt.
 * The LineSplitter class is used to separate lines in line-delimited JSON, XML or TEXT files.
 * It should also work with gzip-compressed line-delimited files.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
llinggit: added below lines
     * Used to return the number of objects in the stream. 
     * @return the number of objects in the stream.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/LineSplitter.java
llinggit: It should work with gzipped XML File. I changed the description as below.

     * Takes the input stream and converts it into a stream of StringHandle. The content could be
     * line-delimited JSON, XML or TEXT file. It could also be gzip-compressed line-delimited files.
     * Provide GZIPInputStream to the splitter when splitting gzip files.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
llinggit: I just realized that I don't need a connection to the database. I only checked the StringHandles returned by the splitter. I'm removing the connection part.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
llinggit: I removed the After part as there's no need for a connection.

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
llinggit: That's true! Thanks!

marklogic-client-api/src/test/java/com/marklogic/client/test/LineSplitterTest.java
llinggit: added below method
private void checkContent(Stream<StringHandle> contentStream,
                                 Format format,
                                 String[] originalResult) {...}

PR #1167, opened 2019-11-25
kafka source connector and tasks with the config and endpoints .

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-94
* Version of MarkLogic Java Client API - 5
* OS and version

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/MarklogicSourceConnector.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/MarklogicSourceConnector.java
@@ -1,0 +1,62 @@
+package com.marklogic.client.test.dataservices;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.source.SourceConnector;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MarklogicSourceConnector extends SourceConnector {
+    private String filename;
+    String FILE_CONFIG = "file_config";
+    String TOPIC_CONFIG = "topic_config";
+    private Map<String, String> config;
+    private String topic;
+    @Override
+    public void start(Map<String, String> props) {
+       // filename = props.get(FILE_CONFIG);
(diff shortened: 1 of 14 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/MarklogicSourceConnector.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public List<Map<String, String>> taskConfigs(int maxTasks) {
+// number of tasks = number of forests
+        // forestname passed to tasks and tasks puts it in workunit
+        // map - "forestname", forest_1 etc
+
+
+        final List<Map<String, String>> configs = new ArrayList<>(3);
+        for (int i = 0; i < 3; i++) {
+            config.put("forestName","java-unittest-"+(i+1));
```
ehennum: Shouldn't each item in the list be a different Map?  If it's the same Map, won't the last put() to "forestName" be the assignment for all tasks?  

Would it make sense to create a new Map within the loop, copy the properties with putAll(), and then set the "forestName"?

marklogic-client-api/src/test/java/com/marklogic/client/test/dataservices/MarklogicSourceTask.java
```diff
@@ -0,0 +1,91 @@
...
+            apiName = props.get("apiName");
+            apiObj = IOTestUtil.readApi(apiName);
+            scriptPath = IOTestUtil.getScriptPath(apiObj);
+            apiPath = IOTestUtil.getApiPath(scriptPath);
+            IOTestUtil.load(apiName, apiObj, scriptPath, apiPath);
+            String forestName = props.get("forestName");
+           String endpointState = "{\"next\":"+1+"}";
+            String workUnit      = "{\"limit\":"+100+"}"+"{\"forestName\":"+forestName+"}";
+            bulkCaller = OutputEndpoint.on(IOTestUtil.db, new JacksonHandle(apiObj)).bulkCaller();
+            bulkCaller.setEndpointState(new ByteArrayInputStream(endpointState.getBytes()));
```
ehennum: One simpler alternative might be to use StringHandle so there's no need to get the bytes from the strings.

PR #1169, opened 2019-12-03
Add class and unit test for DHFPROD-3136.

llinggit:
* What issue are you addressing with this pull request? - DHFPROD-3136
* Are you modifying the correct branch? (See CONTRIBUTING.md) - develop
* Have you run unit tests? (See CONTRIBUTING.md) - yes
* Version of MarkLogic Java Client API (see Readme.txt) - 5.0
* Java version (`java -version`) - 11
* OS and version - MacOS
ZipSplitter


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Returns the entryFileter set to splitter.
+     * @return the entryFileter set to splitter
+     */
+    public Predicate<ZipEntry> getEntryFilter() {
+        return this.entryFilter;
+    }
+
+    /**
+     * Used to set entryFilter to splitter.
```
ehennum: Could the JavaDoc tell the developer that they can use a lambda to inspect the zip entry and return false for any document in the zipfile that should be ignored?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Returns the uriTransformer set to splitter.
+     * @return the uriTransformer set to splitter
+     */
+    public Function<String, String> getUriTransformer() {
+        return this.uriTransformer;
+    }
+
+    /**
+     * Used to set uriTransformer to splitter
```
ehennum: Could the JavaDoc tell the developer that they can use a lambda to transform the name of the document in the zipfile into the document URI for the database?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+ */
+public class ZipSplitter implements Splitter<BytesHandle> {
+    private Map<String, Format> extensionFormats;
+    private Predicate<ZipEntry> entryFilter;
+    private Function<String, String> uriTransformer;
+    private int count = 0;
+
+    /**
+     * Returns the extensionFormats set to splitter. The extensionFormat is pre-defined in splitter.
+     * It includes "json", "txt" and "xml" extensions. If the file has no extension, it is treated as binary file.
```
ehennum: Could the JavaDoc also let the developer know that she can add mappings from other extensions of documents in the zipfile to one of the four MarkLogic formats?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    @Override
+    public long getCount() {
+        return this.count;
+    }
+
+    /**
+     * Takes a input stream of a ZIP file and convert it to a stream of BytesHandle.
+     * The ZIP file could contain XML, JSON, TXT and BINARY files.
```
ehennum: Could the JavaDoc let the developer know that the input stream must be a ZipInputStream?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+            throw new IllegalArgumentException("Input cannot be null");
+        }
+
+        BytesHandleSpliterator bytesHandleSpliterator = new BytesHandleSpliterator(this);
+        bytesHandleSpliterator.setZipStream(input);
+        bytesHandleSpliterator.setEntryFilter(this.entryFilter);
+        bytesHandleSpliterator.setExtensionFormats(this.extensionFormats);
+        bytesHandleSpliterator.setUriTransformer(this.uriTransformer);
+
+        return StreamSupport.stream(bytesHandleSpliterator, false);
```
ehennum: Given our decision that the order of documents within the stream isn't important, maybe the second argument (whether the created stream should support parallel processing of the produced byte handles) should be true.

What do you think?

Just to be clear, the same question applies to the other call to StreamSupport.stream() in splitWriteOperations()

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+                if (nextEntry == null) {
+                    return false;
+                }
+
+                BytesHandle nextBytesHandle = readEntry(nextEntry);
+                action.accept(nextBytesHandle);
+                splitter.count++;
+
+            } catch (IOException e) {
+                e.printStackTrace();
```
ehennum: The IDEs tend to generate boilerplate code that suppresses exceptions.

Unless we know it is safe to suppress an exception, the best practice is to throw the exception. In this case, if we encounter an IOException, something must be wrong with the zipfile.

Where the exception is a checked exception (like IOException) that's not allowed by the signature (as with the tryAdvance() method), we can rethrow it with something like

    throw new RuntimeException("Could not read ZipEntry", e);

To be clear, the same issue affects the other tryAdvance() method.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+
+            while ((candidateEntry = getZipStream().getNextEntry()) != null) {
+                if (getEntryFilter() != null && getEntryFilter().test(candidateEntry) == false) {
+                    continue;
+                }
+
+                String name = candidateEntry.getName();
+                if (getUriTransformer() != null) {
+                    name = getUriTransformer().apply(name);
+                }
```
ehennum: My mistake.  DocumentWriteOperationSpliterator.tryAdvance() calls the URI transformer in the correct place.

I don't think we should call URI transformer here -- that would be calling it twice on the same name.

Maybe the uriTransformer field and its getter and setter methods should move from the base ZipEntrySpliterator class to the derived DocumentWriteOperationSpliterator class.  If so, we should only initialize the DocumentWriteOperationSpliterator instance with the uriTransformer and not the BytesHandleSpliterator instance.

Apologies for the mistaken suggestion during the working session.

marklogic-client-api/src/test/java/com/marklogic/client/test/ZipSplitterTest.java
```diff
@@ -0,0 +1,72 @@
...
+            DocumentWriteOperation docOp = itr.next();
+            assertNotNull(docOp.getUri());
+            assertEquals(docOp.getUri(), zipEntry.getName());
+
+            assertNotNull(docOp.getContent());
+            String docOpContent = docOp.getContent().toString();
+            checkContent(zipInputStream, zipEntry, docOpContent);
+        }
+    }
+
```
ehennum: Should we add a filter lambda test on split() and a uriTransformer test on splitWriteOperations()?

Just to make sure that the lambdas are wired in correctly?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -0,0 +1,363 @@
...
+     * It includes "json", "txt" and "xml" extensions. If the file has no extension, it is treated as binary file.
+     * @return a map of extensionFormats
+     */
+    public Map<String, Format> getExtensionFormats() {
+        return this.extensionFormats;
+    }
+
+    /**
+     * Returns the entryFileter set to splitter.
+     * @return the entryFileter set to splitter
```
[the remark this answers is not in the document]
llinggit: Fixed the typo.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Added following description. 
"The entryFilter is a lambda function, which can be used to inspect the zip entry and return false for any document in the zip file that should be ignored. "

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Added below description.
"The uriTransformer is a lambda function, which can be used to transform the name of the document in the zipfile into the document URI for the database."

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Add below description.
"You can also add mappings from other extensions in the zipfile to one of the four MarkLogic formats."

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Added below description.
"The input stream must be a ZipInputStream, otherwise it will throw an exception."

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: I agree. I could set the second parameter to true to enable parallel processing. Then should we add Splitorator.CONCURRENT to BytesHandleSpliterator as well?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Replaced "e.printStackTrace();" with "throw new RuntimeException("Could not read ZipEntry", e);"

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
ehennum: As I read the JavaDoc for Spliterator.CONCURRENT ("Characteristic value signifying that the element source may be safely concurrently modified"), it is about concurrent modification of an item rather than concurrent processing of items in the stream.

If that reading is correct, we CONCURRENT probably isn't appropriate, but the question is entirely appropriate.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Removed getUriTransformer call in getNextEntry method.
Removed setUriTransfomer in split method as bytesHandleSpliterator doesn't need it.
Moved getter and setter f uriTransformer field from base ZipEntrySpliterator to DcoumentWriteOperationSplitorator.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Got it. Thank you for the answer.

marklogic-client-api/src/test/java/com/marklogic/client/test/ZipSplitterTest.java
llinggit: Sure. Added an entryFilter to test the split method and a uriTransformer to test splitWriteOperations method.

splitter.setEntryFilter(x -> x.getSize() > 50 ? true : false );
splitter.setUriTransformer(name -> name.toUpperCase());

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
llinggit: Actually we don't need the getter and setter in DcoumentWriteOperationSplitorator. Because we pass in ZipSplitter splitter when constructing DocumentWriteOperationSpliterator, we could call splitter.uriTransformer directly. What do you think?

ehennum: Sounds good to me.

Issue #1168, opened 2019-11-26
sort documents by URI for each WriteBatch


ehennum:
Before writing a batch to the server, the WriteBatcher should sort the documents by URI within the batch.

That way, batches written concurrently cannot deadlock because documents will be locked in a consistent order.  That is, whichever batch gets the lock on the first document that is common between batches will get the locks on any other documents that are common between batches.

ehennum:
This fix should apply not only to DMSDK but to any batch writer.

ehennum:
See internal task Java-159.

georgeajit:
Tests are running fine on Jenkins CI builds.

PR #1174, opened 2019-12-13
sort documents by URI within a batch.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-159
* Version of MarkLogic Java Client API 5.0
* Java version 11.0
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteOperationImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteOperationImpl.java
@@ -30,24 +30,15 @@ public class DocumentWriteOperationImpl implements DocumentWriteOperation,Compar
   public DocumentWriteOperationImpl(OperationType type, String uri,
                                     DocumentMetadataWriteHandle metadata, AbstractWriteHandle content)
   {
-    this(uri, type, metadata, content);
-    this.temporalDocumentURI = null;
+    this(type, uri, metadata, content, null);
   }
 
   public DocumentWriteOperationImpl(OperationType type, String uri,
                                     DocumentMetadataWriteHandle metadata, AbstractWriteHandle content, String temporalDocumentURI) {
-    this(uri,type,metadata, content);
-    this.temporalDocumentURI = temporalDocumentURI;
-  }
-
-  private DocumentWriteOperationImpl(String uri, OperationType type,
-                                     DocumentMetadataWriteHandle metadata, AbstractWriteHandle content) {
-
-
     if(type == OperationType.DOCUMENT_WRITE && uri == null) {
       throw new IllegalArgumentException("Uri cannot be null when Operation Type is DOCUMENT_WRITE.");
     }
-    if(!(type == OperationType.DOCUMENT_WRITE) && uri != null) {
+    if(type != OperationType.DOCUMENT_WRITE && uri != null) {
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteOperationImpl.java
```diff
@@ -70,4 +81,37 @@ public DocumentMetadataWriteHandle getMetadata() {
   public AbstractWriteHandle getContent() {
     return content;
   }
+
+  @Override
+  public int compareTo(DocumentWriteOperation o) {
+    if(o == null)
+      throw new NullPointerException("DocumentWriteOperation cannot be null");
+
+    if(this.getUri() instanceof String && o.getUri() instanceof String)
```
ehennum: instanceof implies that a specific subtype instead of null is being checked.

Would != null be easier to understand relative to the == null tests?

marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteOperationImpl.java
```diff
@@ -70,4 +81,37 @@ public DocumentMetadataWriteHandle getMetadata() {
...
+      return getUri().compareTo(o.getUri());
+
+    if(this.getUri() == null && o.getUri() instanceof String)
+      return -1;
+
+    if(this.getUri() instanceof String && o.getUri()==null)
+      return 1;
+
+    if(this.getUri() == null && o.getUri() == null)
+      super.hashCode();
```
ehennum: compareTo() is defined to return -1, 0, or 1.  

super.hashCode() will rarely be one of these 3 numbers.

Would super.compareTo(o) provide a way to fall back to the default object collation?

marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteOperationImpl.java
```diff
@@ -70,4 +81,37 @@ public DocumentMetadataWriteHandle getMetadata() {
...
+  public int hashCode() {
+    if(this.getUri()!=null)
+      return this.getUri().hashCode();
+    return super.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object o){
+    if(this.getUri()!=null)
+      return this.getUri().equals(o);
```
ehennum: getUri() returns a String, which can never equal a DocumentWriteOperation

Would it make sense to return

- false if o is not an instance of DocumentWriteOperation
- super.equals() on o if this.getUri() is null
- this.getUri().equals() on the uri of o after casting o to a DocumentWriteOperation

marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentWriteSetImpl.java
```diff
@@ -133,4 +141,81 @@ public DocumentWriteSet add(DocumentDescriptor desc, DocumentMetadataWriteHandle
...
+  }
+
+  @Override
+  public <T> T[] toArray(T[] a) {
+    return operations.toArray(a);
+  }
+
+  @Override
+  public boolean add(DocumentWriteOperation documentWriteOperation) {
+    System.out.println(operations);
```
ehennum: Maybe a log statement slipped through?

anu3990:
Fix for https://github.com/marklogic/java-client-api/issues/1168

Issue #1163, opened 2019-11-20
Unexpected char when inserting doc with Ü in uri 


jeremybrunetML:
#1008 # Version of MarkLogic Java Client API
'com.marklogic', name: 'marklogic-client-api', version: '5.0.0'

### Version of MarkLogic Server
10.0-2.1

### Java version
java 12.0.1

### OS and version
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G1012

### Input: Some code to illustrate the problem, preferably in a state that can be independently reproduced on our end
        GenericDocumentManager docMgr = DatabaseClientConfiguration.getInstance().getAdminClient().newDocumentManager();
        DocumentWriteSet writeSet = docMgr.newWriteSet();
        StringHandle handle = new StringHandle("{hello:\"test\"}");
        writeSet.add("/CXXXX_Ü_9999.json", handle);
        docMgr.write(writeSet);

### Expected output: What specifically did you expect to happen?
document created in ML

### Alternatives: What else have you tried, actual/expected?
Same issue with datamovementSDK writeBatcher
Works properly with:
- TextDocumentManager
- curl on /v1/documents
- documentInsert()

ehennum:
We plan to fix this bug in the next version of the Java API, which should be released concurrent with MarkLogic 10.0-3 (currently scheduled for January).

anu3990:
Hi @jeremybrunetML . I was able to duplicate the error and it comes when we use TextDocumentManager as well.

ehennum:
Fixing this bug requires a change to the REST API (tracked with internal bug 53915) as well as the Java API.

On the Java API side, it will be necessary to check the server version to determine the syntax supported for the Content-Disposition header.

ehennum:
Internal bug 53915 is now committed on the trunk, 10, and 9 branches of the REST API.

**IMPORTANT:** Because the Java API fix depends on a REST API fix, the Java API can only support UTF-8 characters in document URIs for server releases starting with 9.0-12 and 10.0-3 

When writing a batch, the Java API should check whether a document URI consists entirely of characters from the US-ASCII character set. 

If so, the Java API should continue quote the document URI in the current approach for backward compatibility. 

If not, the Java API should percent-encode the document URI according to this format:

    filename*=UTF-8''THE_PERCENT_ENCODED_DOCUMENT_URI;

as in

    filename*=UTF-8''%2FCXXXX_%C3%9C_9999.json;

[...]

PR #1175, opened 2019-12-13
Fix to allow uris with different characters.

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1163
* Are you modifying the correct branch? - develop
* Have you run unit tests? -  yes
* Version of MarkLogic Java Client API- 5.0
* Version of MarkLogic Server 10.0.2
* Java version - 11.0
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
@@ -131,6 +131,7 @@ import java.io.Writer;
 import java.math.BigDecimal;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -3569,8 +3570,21 @@ public void postBulkDocuments(
...
+        if(asciiEncoder.canEncode(write.getUri())) {
+            disposition = DISPOSITION_TYPE_ATTACHMENT + "; " +
+                      DISPOSITION_PARAM_FILENAME + "=" + escapeContentDispositionFilename(write.getUri()) + contentDispositionTemporal;
+        }
+        else {
+            try {
+                disposition = DISPOSITION_TYPE_ATTACHMENT + "; " +
+                          DISPOSITION_PARAM_FILENAME + "*=UTF-8''" + URLEncoder.encode(write.getUri(), "UTF-8") + contentDispositionTemporal;
+              } catch (Exception ex) {
+                  throw new IllegalArgumentException("Uri cannot be accepted");
```
ehennum: Would the error be more useful if more specific?  Maybe "Uri cannot be encoded as UFT-8"?

PR #1180, opened 2020-01-08
Adding changes to sort uris.

anu3990:
* OS and version


marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentWriteOperation.java
```diff
@@ -139,6 +139,10 @@ public String getTemporalDocumentURI() {
                 return null;
             }
 
+            @Override
+            public int compareTo(DocumentWriteOperation o) {
+                return getUri().compareTo(o.getUri());
```
ehennum: Is there any risk of a null pointer exception?  That is, could getUri() return null?  

If I recall correctly, the standard implementation of DocumentWriteOperation handles this case.  Could we move that existing logic into a default method of the DocumentWriteOperation interface so neither implementation of the interface has to implement the compareTo() method?

Issue #1153, opened 2019-10-23
Issue Writing Empty Document Properties


Zipunrar:
Hi MarkLogic,

I do have an issue writing document properties with the new Java Client API Version 4.
Here are the details:

### Version of MarkLogic Java Client API
4.2.0

### Version of MarkLogic Server
9.0.8

### Java version
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)

### Input: Some code to illustrate the problem, preferably in a state that can be independently

```
DatabaseClient markLogic = ... // create client connection
XMLDocumentManager manager = markLogic.newXMLDocumentManager();
DocumentMetadataHandle metadata = new DocumentMetadataHandle();
// this won't work
metadata.getProperties().put("foobar", "");
StringHandle data = new StringHandle("<root>Test</root>");
// error is thrown here
manager.write("/test.xml", metadata, data);
markLogic.release();
```

### Actual output: What did you observe? What errors did you see? Can you attach the logs? (Java logs, MarkLogic logs)
The following stacktrace is thrown:
```
java.lang.NullPointerException
	at com.ctc.wstx.sw.BaseStreamWriter.writeCharacters(BaseStreamWriter.java:464)
	at com.marklogic.client.io.DocumentMetadataHandle$ValueSerializer.process(DocumentMetadataHandle.java:917)
	at com.marklogic.client.impl.ValueConverter.convertFromJava(ValueConverter.java:253)
	at com.marklogic.client.impl.ValueConverter.convertFromJava(ValueConverter.java:94)
	at com.marklogic.client.io.DocumentMetadataHandle.sendPropertiesImpl(DocumentMetadataHandle.java:886)
	at com.marklogic.client.io.DocumentMetadataHandle.sendMetadataImpl(DocumentMetadataHandle.java:788)
	at com.marklogic.client.io.DocumentMetadataHandle.write(DocumentMetadataHandle.java:574)
	at com.marklogic.client.impl.StreamingOutputImpl.writeTo(StreamingOutputImpl.java:58)
	at okhttp3.MultipartBody.writeOrCountBytes(MultipartBody.java:173)
	at okhttp3.MultipartBody.writeTo(MultipartBody.java:114)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:72)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[...]

PR #1188, opened 2020-02-21
Adding null check for XMLStreamWriter.writeCharacters().

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1153 and https://bugtrack.marklogic.com/53490
* Are you modifying the correct branch? yes. develop
* Version of MarkLogic Java Client API - 5.1.0
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/example/util/Bootstrapper.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/example/util/Bootstrapper.java
@@ -461,8 +461,7 @@ public class Bootstrapper {
       }
 
       writer.writeStartElement("port");
-      if(String.valueOf(port) != null)
-        writer.writeCharacters(String.valueOf(port));
+      writer.writeCharacters(String.valueOf(port));
       writer.writeEndElement();
 
       writer.writeEndElement();
--- a/marklogic-client-api/src/main/java/com/marklogic/client/example/util/Bootstrapper.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/example/util/Bootstrapper.java
@@ -439,7 +439,8 @@ public class Bootstrapper {
       writer.writeDefaultNamespace("http://marklogic.com/rest-api");
 
       writer.writeStartElement("name");
-      writer.writeCharacters(server);
+      if(server != null)
+        writer.writeCharacters(server);
       writer.writeEndElement();
 
       if (group != null && group.length() > 0) {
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/example/util/Bootstrapper.java
```diff
@@ -459,7 +461,8 @@ public String toXMLString()
       }
 
       writer.writeStartElement("port");
-      writer.writeCharacters(String.valueOf(port));
+      if(String.valueOf(port) != null)
```
ehennum: Given that an int cannot be null, let's skip this check.

marklogic-client-api/src/main/java/com/marklogic/client/impl/DOMWriter.java
```diff
@@ -136,7 +136,8 @@ public void serializeAttributes(NamedNodeMap attributes) throws XMLStreamExcepti
     }
   }
   public void serializeText(Text text) throws XMLStreamException {
-    serializer.writeCharacters(text.getData());
+    if(text != null && text.getData() != null)
+      serializer.writeCharacters(text.getData());
```
ehennum: Checking whether text is null seems properly defensive.  

If text is not null, I believe that text.getData() is guaranteed not to be null, so that part of the check is unnecessary.

marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentMetadataPatchBuilderImpl.java
```diff
@@ -752,7 +757,8 @@ public void write(JSONStringWriter serializer) {
     public void write(XMLOutputSerializer out) throws Exception {
       XMLStreamWriter serializer = out.getSerializer();
       writeStartReplace(out, "/rapi:metadata/rapi:quality", null);
-      serializer.writeCharacters(String.valueOf(quality));
+      if(String.valueOf(quality) != null)
+        serializer.writeCharacters(String.valueOf(quality));
```
ehennum: Given that quality is an int, I believe it is guaranteed not to be null, which makes the test unnecessary.

marklogic-client-api/src/main/java/com/marklogic/client/io/DocumentMetadataHandle.java
```diff
@@ -897,7 +900,8 @@ private void sendPropertiesImpl(final XMLStreamWriter serializer) throws XMLStre
   private void sendQualityImpl(XMLStreamWriter serializer) throws XMLStreamException {
     if ( qualityModified == false ) return;
     serializer.writeStartElement("rapi", "quality", REST_API_NS);
-    serializer.writeCharacters(String.valueOf(getQuality()));
+    if(String.valueOf(getQuality()) != null)
+      serializer.writeCharacters(String.valueOf(getQuality()));
```
ehennum: If quality is an int, we don't need the check.

PR #1194, opened 2020-03-13
RowBatcher - First Draft.

anu3990:
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
@@ -1,0 +1,316 @@
+package com.marklogic.client.datamovement.impl;
+
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.datamovement.*;
+import com.marklogic.client.expression.PlanBuilder;
+import com.marklogic.client.impl.BatchPlanImpl;
+import com.marklogic.client.impl.DatabaseClientImpl;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.io.StringHandle;
+import com.marklogic.client.io.marker.StructureReadHandle;
+import com.marklogic.client.util.RequestParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.List;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class RowBatcherImpl extends BatcherImpl implements RowBatcher {
+
+    private int batchSize;
+    private  int threadCount;
+    private ForestConfiguration forestConfig;
+    private String jobId;
+    private  String jobName;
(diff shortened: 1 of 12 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
```diff
@@ -0,0 +1,312 @@
...
+    private  String jobName;
+    private JobTicket jobTicket;
+    private Calendar jobStartTime;
+    private Calendar jobEndTime;
+    private PlanBuilder.PreparePlan preparePlan;
+    private DatabaseClient client;
+    private AtomicLong lowerBound = new AtomicLong(0);
+    private long maxBatches = Long.MAX_VALUE;
+    private long rowCount;
+    private Long bucketSize;
```
ehennum: I'm wondering whether the bucket size can be an integer -- that is, whether we would ever retrieve more rows in a single request than Integer.MAX_VALUE.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
```diff
@@ -0,0 +1,312 @@
...
+    private String jobId;
+    private  String jobName;
+    private JobTicket jobTicket;
+    private Calendar jobStartTime;
+    private Calendar jobEndTime;
+    private PlanBuilder.PreparePlan preparePlan;
+    private DatabaseClient client;
+    private AtomicLong lowerBound = new AtomicLong(0);
+    private long maxBatches = Long.MAX_VALUE;
+    private long rowCount;
```
ehennum: Should we initialize this variable to 0?

Should we expose a getRowEstimate() in the RowBatcher interface that returns this value? The application might make some decisions based on this value.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
```diff
@@ -0,0 +1,312 @@
...
+    public boolean isStarted() {
+        return started.get();
+    }
+
+    private void getRowCount(){
+
+        JsonNode jn = ((DatabaseClientImpl) client).getServices()
+                .getResource(null, "internal/viewinfo", null, params,
+                        new JacksonHandle())
+                .get();
```
ehennum: The value of rowCount used in the calculation on current line 248 is obtained by the internal/viewinfo request.

This method should get the value of the tableID and rowCount properties from the JsonNode and assign them to private fields of the RowBatcherImpl.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/RowBatcherImpl.java
```diff
@@ -0,0 +1,312 @@
...
+            logger.warn("startJob called more than once");
+            return;
+        }
+        if ( getBatchSize() <= 0 ) {
+            withBatchSize(1);
+            logger.warn("batchSize should be 1 or greater--setting batchSize to 1");
+        }
+        jobTicket = ticket;
+
+        getRowCount();
```
ehennum: Once we know the batch size, we can spin up the threads and submit tasks.

As sketched in the technical story, each task should 

1. get the next lower bound
2. calculate the upper bound
3. bind the parameters on the plan to the lower and upper bound values
4. use the RowManager to get the result rows
5. send the result rows to the success listener

This code will be similar to startQuerying() and QueryTask in QueryBatcherImpl

anu3990:
Closing this request for now. Will be opening a new one with unit tests.

Issue #1197, opened 2020-03-18
Add JSONSplitter


llinggit:
For reference, https://project.marklogic.com/jira/browse/JAVA-165

From Jira task,

"

> Large sets of records are often serialized as a single JSON data structure (typically, an array containing an object for each record).
> 
> For that reason, adopters should be able to split an JSON file into separate payloads for writing to the database.
> 
> Each payload must be buffered because the application might accumulate a batch of payloads before sending the payloads to the server. 
> 
> The design should resemble the design of the XML splitter in that the JSON splitter will walk the data structure tree, notifying a visitor about container nodes so the visitor can decide whether to split out the container node as a separate payload and how to represent that payload (which can matter if the application needs to change or inspect the payload downstream).
> 
> One important difference is that a JSON data structure tree has two kinds of container nodes (arrays and objects) where XML only has one (elements). When an array contains other arrays, the visitor might need to know the depth to decide whether or not to split out the container node as a payload. The developer should track array depth within multidimensional arrays in their visitor. 

"
So we can address your issue, please include the following:
### Version of MarkLogic Java Client API 5
See Readme.txt

### Version of MarkLogic Server 10

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version 8

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

georgeajit:
Tests are written and they are part of all splitter tests.

PR #1198, opened 2020-03-19
Add class and unit test for JAVA-165, #1197

llinggit:
* What issue are you addressing with this pull request? #1197 JSONSplitter
* Version of MarkLogic Java Client API (see Readme.txt) 5
* Java version (`java -version`) 8
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
@@ -139,25 +139,22 @@ public class JSONSplitter<T extends AbstractWriteHandle> implements Splitter<T> 
     }
 
     /**
-     * The Visitor class is used to describe the rule of how to split the JSON file.
+     * The Visitor class is used to accumulate and inspect state during the depth-first traversal of the JSON tree
+     * and make the decision of how to split the JSON file.
      * It checks if the current object or array is the target to split.
      * If it is, convert the target object or array into buffered handles or DocumentWriteOperations.
      * @param <T> The type of the handle used for each split
      */
     static public abstract class Visitor<T extends AbstractWriteHandle>   {
 
-        public int arrayDepth = 0;
-
         /**
-         * Use arrayDepth and containerKey to check if the current object is the one to split.
+         * This method inspects the state of the current object and decides whether to split it or not. This method
+         * can be overridden to accumulate and inspect state during the depth-first traversal of the JSON tree and make
+         * the decision of whether to split.
          * @param containerKey The key of the object which the value contains current object
          * @return different operations to either process current object, go down the JSON tree or skip current object
          */
         public NodeOperation startObject(String containerKey) {
-            if (arrayDepth > 0) {
-                return NodeOperation.PROCESS;
-            }
-
             return NodeOperation.DESCEND;
         }
 
@@ -169,18 +166,13 @@ public class JSONSplitter<T extends AbstractWriteHandle> implements Splitter<T> 
         }
 
         /**
-         * Use the arrayDepth and containerKey to check if the current array is the one to split. Also increase arrayDepth.
+         * This method inspects the state of the current array and decides whether to split it or not. This method
+         * can be overridden to accumulate and inspect state during the depth-first traversal of the JSON tree and make
+         * the decision of whether to split.
          * @param containerKey The key of the object which the value contains current array
          * @return different operations to either process current array, go down the JSON tree or skip current array
          */
         public NodeOperation startArray(String containerKey) {
-            arrayDepth++;
-
-            if (arrayDepth > 1) {
-                arrayDepth--;
-                return NodeOperation.PROCESS;
-            }
-
             return NodeOperation.DESCEND;
         }
 
@@ -267,10 +259,10 @@ public class JSONSplitter<T extends AbstractWriteHandle> implements Splitter<T> 
          * @return different operations to either process current array or go down the JSON tree
          */
         public NodeOperation startArray(String containerKey) {
-            arrayDepth++;
+            incrementArrayDepth();
 
             if (arrayDepth > 1) {
-                arrayDepth--;
+                decrementArrayDepth();;
                 return NodeOperation.PROCESS;
             }
 
@@ -282,6 +274,28 @@ public class JSONSplitter<T extends AbstractWriteHandle> implements Splitter<T> 
          * @param containerKey The key of the object which the value contains current array
          */
         public void endArray(String containerKey) {
+            decrementArrayDepth();
+        }
+
+        /**
+         * Get the current array depth in the JSON tree.
+         * @return current array depth
+         */
+        public int getArrayDepth() {
+            return arrayDepth;
+        }
+
+        /**
+         * Increment array depth by 1 while traversing the JSON tree.
+         */
+        public void incrementArrayDepth() {
+            arrayDepth++;
+        }
+
+        /**
+         * Decrement array depth by 1 while traversing the JSON tree.
+         */
+        public void decrementArrayDepth() {
             arrayDepth--;
         }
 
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
```diff
@@ -0,0 +1,550 @@
...
+                        case FIELD_NAME:
+                            key.pop();
+                            key.push(jsonParser.getCurrentName());
+                            break;
+
+                        case START_OBJECT:
+                            NodeOperation operation = visitor.startObject(key.peek());
+                            switch (operation) {
+                                case DESCEND:
+                                    key.push(currentToken.asString());
```
ehennum: Should the START_OBJECT case always push a string by moving line 351 before line 348, given that the END_OBJECT case always pops a key?

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
```diff
@@ -0,0 +1,550 @@
...
+        }
+
+        private void maintainDepth(JsonToken currentToken) {
+            if (currentToken == JsonToken.START_ARRAY || currentToken == JsonToken.START_OBJECT) {
+                depth++;
+            }
+
+            if (currentToken == JsonToken.END_ARRAY || currentToken == JsonToken.END_OBJECT) {
+                depth--;
+            }
```
ehennum: Would a switch be easier to understand?  The two cases that should be treated the same way can just fall through, as in:

```
switch(currentToken) {
    case START_ARRAY:
    case START_OBJECT:
        depth++;
        break;
...
} 
```

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
```diff
@@ -0,0 +1,550 @@
...
+            maintainDepth(super.currentToken());
+        }
+
+        @Override
+        public JsonToken nextToken() throws IOException {
+            if (depth == 0) {
+                throw new IllegalStateException("The JSON branch is closed");
+            }
+
+            JsonToken next = super.nextToken();
```
ehennum: Would it be simpler to call the maintainDepth() method that takes no arguments?  Or even to combine the maintainDepth() methods as a method that takes no arguments and has line 490 as its first statement?

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
```diff
@@ -0,0 +1,241 @@
...
+        };
+
+        for (int i = 0; i < result.length; i++) {
+            String element = result[i].get();
+            assertNotNull(element);
+            assertEquals(expected[i], element);
+        }
+    }
+
+    static public class MultiArrayVisitor extends JSONSplitter.Visitor {
```
ehennum: Should this test derive from ArrayVisitor? Should ArrayVisitor refactor the increment and decrement operations into public incrementArrayDepth() and decrementArrayDepth() methods as conveniences for subclasses that need to manage the array depth?

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
```diff
@@ -0,0 +1,241 @@
...
+        };
+
+        for (int i = 0; i < result.length; i++) {
+            String element = result[i].get();
+            assertNotNull(element);
+            assertEquals(expected[i], element);
+        }
+    }
+
+    static public class KeyVisitor extends JSONSplitter.Visitor {
```
ehennum: Should this test subclass ArrayVisitor if it needs to manage array depth?

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
```diff
@@ -0,0 +1,241 @@
+package com.marklogic.client.test.datamovement;
```
ehennum: Because the test will be in a public repository, please add the copyright header.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
```diff
@@ -0,0 +1,550 @@
...
+
+        public int arrayDepth = 0;
+
+        /**
+         * Use arrayDepth and containerKey to check if the current object is the one to split.
+         * @param containerKey The key of the object which the value contains current object
+         * @return different operations to either process current object, go down the JSON tree or skip current object
+         */
+        public NodeOperation startObject(String containerKey) {
+            if (arrayDepth > 0) {
```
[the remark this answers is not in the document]
llinggit: Go it. 
1. I removed line 149, the declaration of public int arrayDepth in Visitor class.
2. I changed startObject and startArray in Visitor class to only return NodeOpertaion.DESCEND.
3. Added private arrayDepth and override methods startObject and endArray in MultiArrayVisitor in the test and tests pass.
4.  Changed JavaDoc of startObject and startArray in Visitor class to text like below.

> This method inspects the state of the current object and decides whether to split it or not. This method
>          * can be overridden to accumulate and inspect state during the depth-first traversal of the JSON tree and make
>          * the decision of whether to split.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
llinggit: No, because only in case DESCEND, we go down into this object and will reach END_OBJECT. In two other cases, PROCESS and SKIP, we won't reach END_OBJECT to pop out the key.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
llinggit: Yes, changed the if/else clause to switch clause.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
llinggit: I changed maintainDepth() as below

```
private void maintainDepth() throws IOException {
            JsonToken next = super.nextToken();
            maintainDepth(next);
        }
```

And changed blow function to use maintainDepth() alone like below
```
@Override
        public Boolean nextBooleanValue() throws IOException {
            if (depth == 0) {
                return null;
            }

            maintainDepth();
            return super.getBooleanValue();
        }
```

I'm also wondering is the logic for isClosed() wrong? Should we return (depth <= 0) instead of (depth > 0). Then when depth is 0, it returns true.

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
llinggit: Added. Thanks.

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
llinggit: Yes, that's better. 

I changed both my Visitor tests to extend ArrayVisitor. Added 3 public methods in ArrayVisitor, getArrayDepth(), incrementArrayDepth() and decrementArrayDepth(). Thank you!

marklogic-client-api/src/test/java/com/marklogic/client/test/datamovement/JSONSplitterTest.java
llinggit: Yes, fixed as in the above comment.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/JSONSplitter.java
ehennum: That makes sense.  Good catch.

llinggit:
Opened a new pull request #1199 to merge in branch develop-165.
https://github.com/marklogic/java-client-api/pull/1199

Issue #1196, opened 2020-03-18
optional forestName parameter on DocumentManager.search()


ehennum:
This task requires declaring a new interface method and passing a new parameter down the call path.

The expectation for this parameter arises from a similar search() overload in QueryManager:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/query/QueryManager.java#L286

The steps of this task:

1.  In the DocumentManager interface, add a new search() overload method that takes all of the existing parameters but with an additional String forestName parameter as the last parameter:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentManager.java#L675

Provide JavaDoc for this new interface method by copying the JavaDoc for an existing search() overload and adding documentation for the forestName parameter that's equivalent to the documentation of the forestName parameter on the similar QueryManager.search() method.

2.  In the DocumentManagerImpl class, modify the private search() method to take an additional String forestName parameter as the last parameter:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentManagerImpl.java#L517

Add the @Override implementation of the new search() overload method after the last existing @Override search() overload method and before the private search() method, returning the result of a call that passes all parameters through to the private search() method.

Modify all calls from the existing @Override search() overload methods to the private search() method to pass a final null value as the forestName parameter.

At current line 541, modify the call to getBulkDocuments() to pass the forestName parameter after the null value:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/DocumentManagerImpl.java#L541

3. In the RESTServices interface, modify the getBulkDocuments() interface method to add a String forestName parameter as the last parameter:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/RESTServices.java#L147

4.  In the OkHttpServices class, modify the @Override getBulkDocuments() implementation method to add a String forestName parameter as the last parameter:

https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java#L945

At current line 957, modify the call to getBulkDocumentsImpl() to add the forestName parameter as the last parameter:

[...]

PR #1201, opened 2020-03-25
code and unit test for #1196

llinggit:
* What issue are you addressing with this pull request? Add forestName to DocumentManager search method
* Are you modifying the correct branch? (See CONTRIBUTING.md) Yes, develop branch
* Version of MarkLogic Java Client API (see Readme.txt) 5.1
* Version of MarkLogic Server (see admin gui on port 8001) 10.0-2
* OS and version Mac

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentManager.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentManager.java
@@ -594,7 +594,7 @@ public interface DocumentManager<R extends AbstractReadHandle, W extends Abstrac
   DocumentPage search(QueryDefinition querydef, long start);
 
   /**
-   * Just like {@link QueryManager#search(QueryDefinition, SearchReadHandle, long) QueryManager.search}
+   * Just like {@link QueryManager#search(QueryDefinition, SearchReadHandle, long, String) QueryManager.search}
    * but return complete documents via iterable DocumentPage.  Retrieves up to getPageLength()
    * documents in each DocumentPage. If setMetadataCategories has
    * been called, populates metadata for each result in the format specified by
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentManager.java
```diff
@@ -593,6 +593,19 @@ DocumentDescriptor exists(String docId, Transaction transaction)
    */
   DocumentPage search(QueryDefinition querydef, long start);
 
+  /**
+   * Just like {@link QueryManager#search(QueryDefinition, SearchReadHandle, long) QueryManager.search}
```
ehennum: Should the link both here and in the other new methods go to the equivalent overload of QueryManager.search() that has the same arguments including the forestName?

marklogic-client-api/src/test/java/com/marklogic/client/test/document/DocumentWriteOperationTest.java
```diff
@@ -309,6 +308,56 @@ public void DocumentWriteSetWithDisableMetadataTest_2() {
...
+            StringHandle handle = new StringHandle("Document - " + i).withFormat(Format.TEXT);
+            batch.add(uri, handle);
+        }
+
+        textDocumentManager.write(batch);
+        StructuredQueryDefinition query = new StructuredQueryBuilder().collection(collectionName);
+
+        DocumentPage documentsF1 = textDocumentManager.search(query, 1, "java-unittest-1");
+        DocumentPage documentsF2 = textDocumentManager.search(query, 1, "java-unittest-2");
+        DocumentPage documentsF3 = textDocumentManager.search(query, 1, "java-unittest-3");
```
ehennum: It might be slightly easier to maintain if the test 

- declared a count on the number of forests
- instantiated a DocumentPage array with the count as length
- iterated from 0 to i < count to initialize each array item with a search result, sum the results, and construct sets

Given that it's a test, however, the change isn't mandatory.

marklogic-client-api/src/main/java/com/marklogic/client/document/DocumentManager.java
llinggit: Yes. Changed those links

marklogic-client-api/src/test/java/com/marklogic/client/test/document/DocumentWriteOperationTest.java
llinggit: Yes, that's cleaner! Changed to below code:
```
int forestCount = 3;
        DocumentPage[] documents = new DocumentPage[forestCount];
        ArrayList<Set<String>> sets = new ArrayList<Set<String>>();

        long totalCount = 0;

        for (int i = 0; i < forestCount; i++) {
            documents[i] = textDocumentManager.search(query, 1, "java-unittest-" + String.valueOf(i+1));
            totalCount += documents[i].getTotalSize();
            sets.add(new HashSet<String>());
            for (DocumentRecord document : documents[i]) {
                sets.get(i).add(document.getUri());
            }

        }
        assertEquals(30, totalCount);
```

PR #1203, opened 2020-03-27
Refactoring BulkIO to use CallContext.

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-170 and 
https://project.marklogic.com/jira/browse/JAVA-171
* Version of MarkLogic Java Client API - 5.1.0
* Java version - 11.0
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
@@ -25,7 +25,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.InputStream;
 import java.util.concurrent.Callable;
-import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
 
 final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoint {
@@ -103,11 +102,11 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
         private ErrorListener errorListener;
 
         private BulkExecCallerImpl(ExecEndpointImpl endpoint, CallContext callContext) {
-            super(endpoint, callContext);
+            super(callContext);
             this.endpoint = endpoint;
         }
         private BulkExecCallerImpl(ExecEndpointImpl endpoint, CallContext[] callContexts, int threadCount) {
-            super(endpoint, callContexts, threadCount, threadCount);
+            super(callContexts, threadCount, threadCount);
             this.endpoint = endpoint;
         }
 
@@ -120,14 +119,18 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
             setPhase(WorkPhase.RUNNING);
             logger.trace("exec running endpoint={} work={}", getEndpointPath(), getCallContext().getWorkUnit());
 
-            if(getThreadCount() == 1)
-                processOutput(getCallContext());
+            if(getThreadCount() == 1 && getCallContexts().size()==1)
+                processOutput();
+            else if(getThreadCount() == 1 && getCallContexts().size()>1) {
+                while(!getCallContexts().isEmpty()) {
+                    processOutput();
+                }
+            }
             else {
                 try {
                     for (int i = 0; i < getThreadCount(); i++) {
                         BulkCallableImpl bulkCallableImpl = new BulkCallableImpl(this);
-                        submit(bulkCallableImpl);
-
+                        submitTask(bulkCallableImpl);
                     }
                     getCallerThreadPoolExecutor().awaitTermination(Long.MAX_VALUE, TimeUnit.DAYS);
                 }
@@ -142,14 +145,17 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
             this.errorListener = errorListener;
         }
 
-        /*private void processOutput() {
-            calling:while (true){
-            processOutput(getCallContext());
+        private void processOutput() {
+            CallContext callContext = getCallContexts().poll();
+            if(callContext != null) {
+                calling:while (true){
+                    if(!processOutput(callContext))
+                        break calling;
+                }
             }
-        }*/
+        }
 
-        private void processOutput(CallContext callContext){
-            calling: while (true) {
+        private Boolean processOutput(CallContext callContext){
                 InputStream output = null;
                 try {
                     logger.trace("exec calling endpoint={} count={} state={}",
@@ -176,21 +182,20 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
                         setPhase(WorkPhase.INTERRUPTED);
                         logger.info("exec interrupted endpoint={} count={} work={}",
                                 getEndpointPath(), getCallCount(), callContext.getWorkUnit());
-                        break calling;
+                        return false;
                     case RUNNING:
                         if (output == null) {
                             setPhase(WorkPhase.COMPLETED);
                             logger.info("exec completed endpoint={} count={} work={}",
                                     getEndpointPath(), getCallCount(), callContext.getWorkUnit());
-                            break calling;
+                            return false;
                         }
-                        break;
+                        return true;
                     default:
                         throw new MarkLogicInternalException(
                                 "unexpected state for "+getEndpointPath()+" during loop: "+getPhase().name()
                         );
                 }
-            }
         }
 
         private class BulkCallableImpl implements Callable<Boolean> {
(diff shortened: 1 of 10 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -119,21 +123,23 @@ public void checkAllowedArgs(InputStream endpointState, SessionState session, In
...
         BulkIOEndpointCallerImpl(IOEndpointImpl endpoint) {
             if (endpoint == null)
                 throw new IllegalArgumentException("null endpoint definition");
+            this.callContext = new CallContextImpl(endpoint);
             this.endpoint = endpoint;
         }
 
         private IOEndpointImpl getEndpoint() {
-            return this.endpoint;
+            return callContext.getEndpoint();
```
ehennum: See the comment below about the CallContext constructor.  I'm wondering whether it might be beneficial to access the endpoint in the field instead of in the call context.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -119,21 +123,23 @@ public void checkAllowedArgs(InputStream endpointState, SessionState session, In
...
         }
 
         String getEndpointPath() {
-            return getEndpoint().getEndpointPath();
+            return callContext.getEndpoint().getEndpointPath();
```
ehennum: Would it be better to use the getEndpoint() getter and thus maintain the access to the endpoint in one place?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -50,17 +50,48 @@ private int getBatchSize() {
...
 	}
+
 	@Override
+	@Deprecated
 	public InputStream call(InputStream endpointState, SessionState session, InputStream workUnit, InputStream[] input) {
-		checkAllowedArgs(endpointState, session, workUnit);
-		return getCaller().arrayCall(getClient(), endpointState, session, workUnit, input);
+		newCallContext().withEndpointState(endpointState).withSessionState(session)
+				.withWorkUnit(workUnit);
+		return call(newCallContext(), input);
```
ehennum: The call on line 61 should use the call context constructed on line 59.

Currently, the constructed call context is thrown away, so the parameters aren't passed in the call.

The equivalent refactoring in InputOutputEndpointImpl is correct.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -103,7 +115,38 @@ private ExecEndpointImpl getEndpoint() {
         @Override
         public void awaitCompletion() {
             setPhase(WorkPhase.RUNNING);
-            logger.trace("exec running endpoint={} work={}", getEndpointPath(), callContext.getWorkUnit());
+            logger.trace("exec running endpoint={} work={}", getEndpointPath(), getCallContext().getWorkUnit());
+
+            if(threadCount == 1)
+                processOutput(getCallContext());
```
ehennum, line 124: After we handle 

*  the simple case of a single call context and single thread
*  the case of multiple call contexts and multiple threads

We'll also need to deal with the case where there are multiple call contexts and a single thread.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -73,28 +76,37 @@ public BulkExecCaller bulkCaller(CallContext callContext) {
 
     @Override
     public BulkExecCaller bulkCaller(CallContext[] callContexts) {
-        return null;
+        if(callContexts == null || callContexts.length==0)
+            throw new IllegalArgumentException("CallContext cannot be null or empty");
+        return bulkCaller(callContexts, callContexts.length);
```
ehennum, line 81: This review doesn't repeat the comments from the review of BulkExecCallerImpl earlier today.

This review assumes that the comments from the earlier review that apply to the other bulk callers don't have to be repeated on those bulk callers.  For one example, the comment on having a single thread count field in the base BulkIOCallerImpl applies generally.  For another example, the feedback on processInput() and on submitting the callable after putting the call context is similar to the comments on processOutput() and the callable in BulkExecCallerImpl.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -137,20 +133,24 @@ public void checkAllowedArgs(InputStream endpointState, SessionState session, In
...
         private CallContext callContext;
-
         private long           callCount = 0;
 
+        private CallerThreadPoolExecutor callerThreadPoolExecutor;
+        private LinkedBlockingQueue<CallContext> callContexts;
+
         BulkIOEndpointCallerImpl(IOEndpointImpl endpoint, CallContext callContext) {
             if (endpoint == null)
                 throw new IllegalArgumentException("null endpoint definition");
```
ehennum, line 146: Why is the BulkIOEndpointCallerImpl() constructor taking an endpoint if it's not doing anything with it?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -327,6 +344,25 @@ void assignEndpointState(InputStream[] output) {
             }
             return output;
         }
+
+        private void checkCallContextQueue() {
+            if(this.callContexts != null || !this.callContexts.isEmpty())
+                throw new InternalError("CallContext queue not empty");
```
ehennum, line 355: Because the error condition that we're preventing is a null pointer, would the check be better as a test that this.callContext != null?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/IOEndpointImpl.java
```diff
@@ -327,6 +344,25 @@ void assignEndpointState(InputStream[] output) {
...
+
+            CallerThreadPoolExecutor(int threadCount, int queueSize) {
+
+                super(threadCount, threadCount, 0, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(queueSize),
+                        new CallerRunsPolicy());
+            }
+
+            private void submitTask(Callable<Boolean> callable) {
+                FutureTask futureTask = new FutureTask(callable);
+                super.execute((Runnable) callable);
```
ehennum, line 368: A Callable is not a Runnable, so this cast will throw a runtime error.

The FutureTask is a Runnable -- that's the reason for wrapping the callable in a FutureTask -- but the FutureTask is thrown away.

The task was suggesting to provide the submit method in the bulk caller instead of the ThreadPoolExecutor.
The submit() method in OutputEndpointImpl.BulkOutputCallerImpl looks correct. On seeing the BulkOutputCallerImpl.submit() method, I'm wondering whether a submit(Callable<Boolean> callable) method could be provided by BulkIOEndpointCallerImpl with package availability so the subclasses don't have to implement submit().

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputEndpointImpl.java
```diff
@@ -208,24 +214,22 @@ private void processInput(CallContext callContext) {
...
+						bulkInputCallerImpl.processInput(callContext);
 
-					bulkInputCallerImpl.processInput(callContext);
 					if(getPhase() == WorkPhase.COMPLETED && bulkInputCallerImpl.getCallContexts().isEmpty()) {
 						getCallerThreadPoolExecutor().shutdown();
-					} else {
+					}
+					else if(callContext != null) {
 						bulkInputCallerImpl.getCallContexts().put(callContext);
+						submitTask(this);
```
ehennum, line 225: By contrast with the output case, the input case is driven by the application adding input.

So, the callable shouldn't resubmit itself but, instead, should shutdown the thread pool if the thread pool is awaiting termination and the task queue is empty.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/InputOutputEndpointImpl.java
```diff
@@ -174,17 +171,24 @@ public void setErrorListener(ErrorListener errorListener) {
         }
 
         private void processInput() {
```
ehennum, line 173: Same comments for input-output case as for input case

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -152,20 +152,25 @@ public void awaitCompletion() {
             }
 
             setPhase(WorkPhase.RUNNING);
```
ehennum, line 154: Same comments for output case as for exec case.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/OutputEndpointImpl.java
```diff
@@ -182,8 +187,17 @@ public void awaitCompletion() {
...
+            CallContext callContext = getCallContexts().poll();
+            if(callContext != null) {
+                calling:while (true){
+                    if(!processOutput(callContext))
+                        break calling;
+                }
+            }
+        }
+
+        private Boolean processOutput(CallContext callContext){
```
ehennum, line 200: Could the return value be the lower-case primitive boolean instead of the Boolean object?  Is there a reason to autobox?

anu3990:
Opening a new pull request - https://github.com/marklogic/java-client-api/pull/1207

Issue #1223, opened 2020-05-18
Add PathSplitter


llinggit:
Please see 
file system splitter - https://project.marklogic.com/jira/browse/JAVA-183
refactor Splitters for DocumentWriteOperation streams - https://project.marklogic.com/jira/browse/JAVA-185

> In one common case, the content to be ingested by DMSDK resides in the local file system.
> This task adds a convenience for DMSDK splitting over a stream of Path objects that identify files in the file system.
> This convenience can provide the foundation for a command-line utility that wraps DMSDK in the future.

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API 
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

PR #1226, opened 2020-05-29
Add Class and test for #1223 #1197 in develop branch

llinggit:
* What issue are you addressing with this pull request? JSONSplitter and PathSplitter
* Are you modifying the correct branch? (See CONTRIBUTING.md) Yes, develop
* Version of MarkLogic Java Client API (see Readme.txt) 5.2
* OS and version Mac


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/PathSplitter.java
```diff
@@ -0,0 +1,193 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.marklogic.client.datamovement;
+
+import com.marklogic.client.document.DocumentWriteOperation;
+import com.marklogic.client.io.marker.AbstractWriteHandle;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
```
ehennum, line 22: We should avoid incurring a compile-time dependency on org.jetbrains.annotations.*

The annotations are useful, but we want to be able to edit the sources in any IDE and also want to minimize our dependencies as much as possible.

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/ZipSplitter.java
```diff
@@ -325,11 +380,25 @@ public boolean tryAdvance(Consumer<? super DocumentWriteOperation> action) {
...
+                }
+
+                if (splitter.getUriMaker() != null) {
+                    if (splitter.inputName != null) {
+                        splitter.getUriMaker().setInputName(splitter.inputName);
+                    }
+                    uri = splitter.getUriMaker().makeUri(splitter.count, name, nextBytesHandle);
+                }
+
+                if (splitter.getUriTransformer() != null) {
```
ehennum, line 401: For greater clarity, this could probably be an else if so execution skips this branch if getUriMaker() is not null

marklogic-client-api/src/main/java/com/marklogic/client/datamovement/PathSplitter.java
llinggit: Got it. Removed the dependency.

Issue #1252, opened 2020-08-28
Error Listener for Bulk Callers


llinggit:
https://project.marklogic.com/jira/browse/JAVA-172

Copied from Jira as below.

```
Whether the bulk caller has a single CallContext or multiple CallContexts, if the call returns an error and the consumer has set an error listener, call the error listener with

the retry number for the call (starting at zero)
the Throwable for the call error
the CallContext for the call
any inputs to the call
Based on the returned ErrorDisposition, either

retry the call (and increment the retry number if the call throws another error)
skip the call but continue to use the CallContext
stop all calls
If the error listener throws an error, the bulk caller stops all calls.
```

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

PR #1253, opened 2020-08-28
Error Listener for Bulk Callers #1252

llinggit:
* What issue are you addressing with this pull request? #1252 
* Are you modifying the correct branch? (See CONTRIBUTING.md) develop branch
* Version of MarkLogic Java Client API (see Readme.txt) 5.0
* OS and version MacOS

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
@@ -25,6 +25,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.InputStream;
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicInteger;
 
 final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoint {
     private static Logger logger = LoggerFactory.getLogger(ExecEndpointImpl.class);
@@ -100,6 +101,7 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
     {
         private ExecEndpointImpl endpoint;
         private ErrorListener errorListener;
+        private AtomicInteger aliveCallContextCount;
 
         private BulkExecCallerImpl(ExecEndpointImpl endpoint, CallContext callContext) {
             super(callContext);
@@ -109,6 +111,7 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
         private BulkExecCallerImpl(ExecEndpointImpl endpoint, CallContext[] callContexts, int threadCount) {
             super(callContexts, threadCount, threadCount);
             this.endpoint = endpoint;
+            this.aliveCallContextCount = new AtomicInteger(threadCount);
         }
 
         private ExecEndpointImpl getEndpoint() {
@@ -145,19 +148,21 @@ final public class ExecEndpointImpl extends IOEndpointImpl implements ExecEndpoi
             this.errorListener = errorListener;
         }
 
-        private void processOutput() {
-            CallContextImpl callContext = getCallContext();
-            if(callContext != null) {
-                while (processOutput(callContext));
-            }
+        private ExecEndpoint.BulkExecCaller.ErrorListener getErrorListener() {
+            return this.errorListener;
         }
 
-        private boolean processOutput(CallContextImpl callContext){
-                InputStream output = null;
+        private InputStream processExec(CallContextImpl callContext) {
+            final int DEFAULT_MAX_RETRIES = 10;
+            ErrorDisposition error = ErrorDisposition.RETRY;
+            InputStream output = null;
+
+            for (int retryCount = 0; retryCount < DEFAULT_MAX_RETRIES && error == ErrorDisposition.RETRY; retryCount++) {
+                Throwable throwable = null;
                 try {
                     logger.trace("exec calling endpoint={} count={} state={}",
                             callContext.getEndpoint().getEndpointPath(), getCallCount(), callContext.getEndpointState());
-// TODO: use byte[] for IO internally (and InputStream externally)
+                    // TODO: use byte[] for IO internally (and InputStream externally)
                     output = getEndpoint().getCaller().call(
                             callContext.getClient(), callContext.getEndpointState(), callContext.getSessionState(),
                             callContext.getWorkUnit()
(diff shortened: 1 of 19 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -145,19 +148,21 @@ public void setErrorListener(ErrorListener errorListener) {
...
-                while (processOutput(callContext));
-            }
+        private ExecEndpoint.BulkExecCaller.ErrorListener getErrorListener() {
+            return this.errorListener;
         }
 
-        private boolean processOutput(CallContextImpl callContext){
-                InputStream output = null;
+        private InputStream processExec(CallContextImpl callContext) {
+            final int DEFAULT_MAX_RETRIES = 10;
```
ehennum, line 156: Is there a reason that DEFAULT_MAX_RETRIES is 10 in some callers and 100 in others?

Could the constant be shared across the implementations if declared as a static final package-visible constant of IOEndpointImpl?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -166,32 +171,78 @@ private boolean processOutput(CallContextImpl callContext){
...
                     incrementCallCount();
-                } catch(Throwable throwable) {
+                    return output;
+                } catch(Throwable catchedThrowable) {
                     // TODO: logging
-                    throw new RuntimeException("error while calling "+getEndpoint().getEndpointPath(), throwable);
+                    throwable = catchedThrowable;
                 }
-// TODO -- retry with new session if times out
+                // TODO -- retry with new session if times out
```
ehennum, line 179: We can remove the comment.  I believe there's no error.  The appserver just sees it as a request to create a new session if the session times out.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -166,32 +171,78 @@ private boolean processOutput(CallContextImpl callContext){
...
-                        if (output == null ) {
-                            if(getCallerThreadPoolExecutor() == null || getCallerThreadPoolExecutor().getActiveCount() <= 1)
-                                setPhase(WorkPhase.COMPLETED);
-                            logger.info("exec completed endpoint={} count={} work={}",
-                                    callContext.getEndpoint().getEndpointPath(), getCallCount(), callContext.getWorkUnit());
-                            return false;
+                if (throwable != null) {
+                    if (getErrorListener() == null) {
+                        logger.error("Error while calling " + getEndpoint().getEndpointPath(), throwable);
+                        throw new RuntimeException("Error while calling " + getEndpoint().getEndpointPath(), throwable);
```
ehennum, line 184: I'm wondering if it might be better to stop all calls if there is no error listener.  If we throw a only runtime exception in a thread, it will get thrown away. Also, that would be more consistent with the case where the error listener throws an error.

What do you think?

At the risk of pointing out the obvious. if we make the change, we'd want to make it in all of the callers.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -166,32 +171,78 @@ private boolean processOutput(CallContextImpl callContext){
...
+            InputStream output = processExec(callContext);
+
+            switch(getPhase()) {
+                case INTERRUPTING:
+                    setPhase(WorkPhase.INTERRUPTED);
+                    logger.info("exec interrupted endpoint={} count={} work={}",
+                            callContext.getEndpoint().getEndpointPath(), getCallCount(), callContext.getWorkUnit());
+                    return false;
+                case RUNNING:
+                    if (output == null ) {
```
ehennum, line 233: Would it be better to check `aliveCallContextCount.get() == 0` instead of `getCallerThreadPoolExecutor().getActiveCount() <= 1` here as well?

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
```diff
@@ -212,9 +263,11 @@ public Boolean call() throws InterruptedException{
                     bulkExecCallerImpl.getCallContextQueue().put(callContext);
                     submitTask(this);
                 }
-                else if(bulkExecCallerImpl.getCallContextQueue().isEmpty() &&
-                        getCallerThreadPoolExecutor().getActiveCount() <= 1) {
-                    getCallerThreadPoolExecutor().shutdown();
+                else {
```
ehennum, line 266: I'd suggest moving the decrementAndGet() call into the if() test.  That way, a different thread can't change the value before this thread gets the value, and only one thread will go to zero.  

The same change would go into the Output endpoint.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
llinggit: Sorry, that's a number set for testing. Added below statement in IOEndpointImpl.java

final static int DEFAULT_MAX_RETRIES = 100;

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
llinggit: Got it. Removed it.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
llinggit: Yes, that's more consistent. Now the IF clause is changed as below.

if (getErrorListener() == null) {
                        logger.error("No error listener set. Stop all calls. " + getEndpoint().getEndpointPath(), throwable);
                        error = ErrorDisposition.STOP_ALL_CALLS;
                    }

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
llinggit: Yes. Also changed in OutputEndpointImpl as well.

marklogic-client-api/src/main/java/com/marklogic/client/dataservices/impl/ExecEndpointImpl.java
llinggit: I'm not quite sure about this one. aliveCallContextCount is AtomicInteger. It guarantees that a different thread can't change the value before this thread changes and gets the value.

llinggit: Thanks for the explanation. Changed the statement as below. 

if (aliveCallContextCount.decrementAndGet() == 0) {
                        getCallerThreadPoolExecutor().shutdown();
                    }

Issue #1246, opened 2020-08-11
minor fixup on JavaDoc for splitters


ehennum:
Running the javadoc Gradle task for the marklogic-client-api/build.gradle file (either from the command line or 
by right clicking in the IDE) reports a few warnings about missing param or  desc tags from the splitter classes.

It's a low priority, but when there's a gap between other commitments, that would be good to clean up.  Thanks!

georgeajit:
```getCoung()``` method needs a description in Java docs.

georgeajit:
Fixed. Verified code and pages after running Javadocs tasks.

llinggit:
There's no server-side change in this issue.

Issue #1245, opened 2020-08-07
Update Java docs for makeUri method's handle parameter.


georgeajit:
### Version of MarkLogic Java Client API
develop
### Version of MarkLogic Server

10.0

### Java version
Java 8

### OS and version

N/A

### Input: Some code to illustrate the problem, preferably in a state that can be independently reproduced on our end
```
/**
     * UriMaker which generates URI for each split file
     */
    public interface UriMaker extends Splitter.UriMaker {
        /**
         * Generates URI for each split
         * @param num the count of each split
         * @param handle the handle which contains the content of each split
         * @return the generated URI of current split
         */
        String makeUri(long num, JacksonHandle handle);
    }
```
### Actual output: What did you observe? What errors did you see? Can you attach the logs? (Java logs, MarkLogic logs)
Java docs for ```handle``` parameter can be updated to provide a hint that contents of ```handle``` can be utilized to make a meaningful document URI. For example, use first field (or column; which may be a primary key) in case of a CSV row as document URI.

### Alternatives: What else have you tried, actual/expected?
N/A

georgeajit:
Verified all usages. Java docs specifies required hint.

Issue #1237, opened 2020-07-21
splitWriteOperations​ method's second parameter could have a better name than inputName


georgeajit:
### Version of MarkLogic Java Client API
develop (post 5.2.0 release)
### Version of MarkLogic Server

10.0 nightly release (post 10.0-4 release)

### Java version

Open jdk 8 or 9

### OS and version

Windows /  IDE

### Input: Some code to illustrate the problem, preferably in a state that can be independently reproduced on our end

```splitWriteOperations​(java.io.InputStream input, java.lang.String inputName)```

### Actual output: 
```inputName``` parameter name does not convey the fact that a filename (as a template name for individual line URIs)  is required.

### Expected output: What specifically did you expect to happen?
N/A
### Alternatives: What else have you tried, actual/expected?
N/A

ehennum:
@georgeajit and @llinggit , how about inputFilename as the parameter name?

llinggit:
I think this parameter means more to the extracted file names. We use this name and sometimes the extension of this parameter to construct the output file names. Because this could be different from the input file's name, that's why I think inputName is a little confusing, but I don't have a better name. inputFilename is kind of similar to inputName I think.

ehennum:
Good point.  How about splitFilename on the principle that we're splitting both the input and the filename?

llinggit:
I think this is much better. How about it @georgeajit ?

georgeajit:
That name along with associated Java Docs support will be fine.

georgeajit:
Test corrected to have new names.

PR #1255, opened 2020-09-10
Make a few changes to JavaDoc and tests

llinggit:
* What issue are you addressing with this pull request? #1237; #1245; #1246; Change count from ThreadLocal to long; Fix test expected results after refactoring of ResendableHandle.
* Version of MarkLogic Java Client API (see Readme.txt) up-to-date
* OS and version


ehennum (approved):
My only question is about the reversion from ThreadLocal<Long> to long.

I seem to recall that we discovered a reason that splitters can't be used in multiple threads, but I don't recall the reason, so I wanted to confirm that reversion was deliberate.

Issue #1260, opened 2020-10-09
Add CtsQueryBuilder for search in Java API


llinggit:
Link to the task in Jira: https://project.marklogic.com/jira/browse/JAVA-196

Copy of description in the task:

"Currently, the Java API provides the StructuredQueryBuilder class to make search easy. 

StructuredQueryBuilder, however, generates structured queries for the Search API. The Search API is both inefficient (because the structured queries have to be transformed to cts.query on the server) and limited (because structured queries expose only a subset of the cts.query constructors).

The Java API also provides support for building a cts.query in the PlanBuilder for the Optic API. Currently, however, this builder cannot be used for search because the queries are serialized using the Optic functional AST. The Optic functional AST makes it possible to declare expressions that yield inputs to the cts.query constructors. The /v1/search endpoint, however, doesn't understand the Optic functional AST.

This improvement enhances the Java API with a cts.query builder to make search easy. See also:
https://github.com/marklogic/java-client-api/issues/376
The existing com.marklogic.client.query package adds two new interfaces and an abstract class."

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

PR #1261, opened 2020-10-09
Add CtsQueryBuilder for search in Java API #1260

llinggit:
* What issue are you addressing with this pull request? #1260
* Are you modifying the correct branch? (See CONTRIBUTING.md) Yes, develop
* Version of MarkLogic Java Client API (see Readme.txt) 5.0
* OS and version Mac

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/expression/CtsQueryBuilder.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/expression/CtsQueryBuilder.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2019 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/expression/CtsQueryBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.client.expression;
+
+import com.marklogic.client.io.marker.JSONReadHandle;
+import com.marklogic.client.io.marker.JSONWriteHandle;
+import com.marklogic.client.query.CtsQueryDefinition;
+import com.marklogic.client.type.CtsQueryExpr;
+
```
ehennum, line 22: Not before checkin, but when you get a chance, adding JavaDoc for the public members on this and other public new classes will be important.

To be clear, implementations like AbstractSearchQueryDefinition don't need JavaDoc.

ehennum (approved):
Looks great -- thanks for the economical and thorough wiring on the builder so we can take advantage of it for document search as well as row queries.

marklogic-client-api/src/main/java/com/marklogic/client/expression/CtsQueryBuilder.java
llinggit: I'll add Javadoc in my next check-in.

PR #1308, opened 2021-07-19
Added pom.xml file to try out github's Dependency Graph

rjrudin:
I'd like to get this added to the master branch so that I can see what Insight / Dependency Graph shows in github - in particular, I want to see what the Dependents view shows. Adding it to the develop branch won't help because github only cares about what's on the master branch. And this has to be a Maven pom.xml file instead of a Gradle build.gradle file since the Dependency Graph feature only supports Maven and not Gradle.

### DIFF
--- a/marklogic-client-api/pom.xml
+++ b/marklogic-client-api/pom.xml
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This file was generated via Gradle and is being used primarily for github's Dependency Graph feature.
+It is not intended to be used to build this project.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <!-- This module was also published with a richer model, Gradle metadata,  -->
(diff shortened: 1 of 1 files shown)


llinggit (approved):

rjrudin:
This can be reverted from master once the results of the Dependency Graph are visible.

ehennum (approved):
Thanks for opening the door on this GitHub capability

rjrudin:
I'm using this task to generate the pom file dynamically, which includes a comment noting the purpose of the file:

```
task generatePomForDependencyGraph(dependsOn: "generatePomFileForMainJavaPublication") {
	description = "Prepare for a release by making a copy of the generated pom file in the root directory so that it " +
		"can enable Github's Dependency Graph feature, which does not yet support Gradle"
	doLast {
		def preamble = '<?xml version="1.0" encoding="UTF-8"?>'
		def comment = "<!--\n" +
			"This file was generated via Gradle and is being used primarily for github's Dependency Graph feature.\n" +
			"It is not intended to be used to build this project.\n" +
			"-->"
		def fileText = file("build/publications/mainJava/pom-default.xml").getText()
		file("pom.xml").setText(fileText.replace(preamble, preamble + comment))
	}
}
```

anu3990:
Hi @rjrudin . Should we merge this pull request?

rjrudin:
I'm going to update this soon for our upcoming 5.6.0 release.

Issue #1316, opened 2021-08-31
Expose the 3rd parameter of op.fromSPARQL


llinggit:
This is for bug 56465. The new 3rd parameter is options, which could be dedup or base. Need to expose this parameter in Java Client API.

const op = require('/MarkLogic/optic');
    op.fromSPARQL('PREFIX ad: <http://marklogicsparql.com/addressbook#> '+
                    'SELECT ?firstName '+
                    'WHERE {<#5555> ad:firstName ?firstName .}'
                 , 'sparql', {dedup:'on',base:'http://marklogicsparql.com/id#'})
                .result();

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

ehennum:
I believe that the only thing missing is to write a unit test using PlanBuilder.sparqlOptions()

llinggit:
OK, will add.

llinggit:
When setting duplicate to true, it works, but when setting it to false, it returns 6 results instead of 2.

PlanSparqlOptions options = pb.sparqlOptions().withDeduplicated(false).withBase("http://marklogicsparql.com/id#");

Issue #1315, opened 2021-08-18
Expose op:sample-by function in Java Client API


llinggit:
As task in bugtrack, https://bugtrack.marklogic.com/55089
A sample query as below

xquery version "1.0-ml";
import module namespace op="http://marklogic.com/optic"
     at "/MarkLogic/optic.xqy";
op:from-view("opticUnitTest", "musician") 
              => op:sample-by(map:map()=>map:with("limit", 2))
              => op:result()

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

ehennum:
Besides writing a unit test using PlanBuilder.sampleByOptions(), please consider whether it makes sense to add

- A PlanBuilderBase.AccessPlanBase.sampleBy() method overload that takes the limit as an int.
- A PlanBuilderSubImpl.AccessPlanSubImpl,sampleBy() implementation that constructs the PlanSampleByOptionsImpl and calls the existing method

llinggit:
Thanks, Erik. Will do.

llinggit:
`PlanSampleByOptions options = p.sampleByOptions().withLimit(2);`

The unit test throws an error, "Server Message: OPTIC-INVALARGS: fn.error(null, 'OPTIC-INVALARGS', 'limit must be a positive number: '+length); -- Invalid arguments: limit must be a positive number: 4"

I found that it errors out at line 2731 of optic-impl.sjs. The number we passed in is treated as a string, not a number. Investigating.

llinggit:
The limit in PlanSampleByOptions is indeed a xs:int, but later changed to String in ExportablePlan.

llinggit:
Proposed changes:

Index: marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
[...]

PR #1320, opened 2021-10-05
Add unit tests for #1314, #1315, #1316. Also modify makeMap to work w…

llinggit:
…ith XsIntVal for fromSparql options.

* What issue are you addressing with this pull request? - #1314, #1315, #1316
* Are you modifying the correct branch? (See CONTRIBUTING.md) - develop
* Have you run unit tests? (See CONTRIBUTING.md) - Yes
* Version of MarkLogic Java Client API (see Readme.txt) - nightly
* Version of MarkLogic Server (see admin gui on port 8001) - 10.0-8
* Java version (`java -version`) - 1.8
* OS and version - Mac

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
@@ -560,13 +560,13 @@ public class PlanBuilderSubImpl extends PlanBuilderImpl {
 
     return mapdef;
   }
-  static Map<String,String> makeMap(PlanSampleByOptions options) {
+  static Map<String,XsIntVal> makeMap(PlanSampleByOptions options) {
     if (options == null) {
       return null;
     }
 
     XsIntVal limit = options.getLimit();
-    return (limit == null) ? null : makeMap("limit", limit.toString());
+    return (limit == null) ? null : makeMap("limit", limit);
   }
   static Map<String,String> makeMap(PlanSparqlOptions options) {
     if (options == null) {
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
```diff
@@ -560,13 +560,13 @@ public PlanParamExpr param(XsStringVal name) {
 
     return mapdef;
   }
-  static Map<String,String> makeMap(PlanSampleByOptions options) {
+  static Map<String,XsIntVal> makeMap(PlanSampleByOptions options) {
```
ehennum, line 563: I have concern about the ability to add other options in the future.  Options are typically string values.

If the limit is a string, will the server cast it to a number?  

If so, I'd suggest that String is a better choice for the data type of the map value by serializing the number to a string before assigning to the map.

If the server doesn't cast, maybe the right approach is to use int for now but to switch to string once the server can cast.

ehennum (approved):
One concern about the future flexibility of the implementation of limit.

marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
llinggit: Hi Erik, the server doesn't cast, and I get exceptions because of it, "Server Message: OPTIC-INVALARGS: fn.error(null, 'OPTIC-INVALARGS', 'limit must be a positive number: '+length); -- Invalid arguments: limit must be a positive number: 4"

I have more comments in https://github.com/marklogic/java-client-api/issues/1315.

ehennum: Thanks for clarifying.  I'd agree with the changes for now but think that it would be good to add a bug on the Optic SJS code to try to cast a string to a number if possible so the Java client can be refactored later if other options become necessary.

Issue #1329, opened 2021-11-30
Expose op.columnInfo on Client API


llinggit:
Related server bug: 57080 - Expose op.columnInfo on Client API

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

PR #1330, opened 2021-11-30
#1329 expose op.columnInfo

llinggit:
* What issue are you addressing with this pull request? #1329 expose op.columnInfo
* Version of MarkLogic Java Client API (see Readme.txt) latest
* Version of MarkLogic Server (see admin gui on port 8001) 10.0-8
* Java version (`java -version`) 1.8
* OS and version Mac

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
@@ -256,7 +256,7 @@ public class RowManagerImpl
   }
 
   @Override
-  public <T extends XMLReadHandle> T generateView(Plan plan, String schema, String view, T resultsHandle) {
+  public <T extends XMLReadHandle> T generateView(PlanBuilder.PreparePlan plan, String schema, String view, T resultsHandle) {
     if (resultsHandle == null) {
       throw new IllegalArgumentException("Must specify a handle to generate a view for the plan");
     } else if (schema == null || schema.length() == 0) {
@@ -276,7 +276,7 @@ public class RowManagerImpl
     return services.postResource(requestLogger, "rows", null, params, astHandle, resultsHandle);
   }
   @Override
-  public <T> T generateViewAs(Plan plan, String schema, String view, Class<T> as) {
+  public <T> T generateViewAs(PlanBuilder.PreparePlan plan, String schema, String view, Class<T> as) {
     ContentHandle<T> handle = handleFor(as);
     if (generateView(plan, schema, view, (XMLReadHandle) handle) == null) {
       return null;
@@ -286,7 +286,7 @@ public class RowManagerImpl
   }
 
   @Override
-  public <T extends JSONReadHandle> T columnInfo(Plan plan, T resultsHandle) {
+  public <T extends JSONReadHandle> T columnInfo(PlanBuilder.PreparePlan plan, T resultsHandle) {
     if (resultsHandle == null) {
       throw new IllegalArgumentException("Must specify a handle to generate a view for the plan");
     }
@@ -300,7 +300,7 @@ public class RowManagerImpl
     return services.postResource(requestLogger, "rows", null, params, astHandle, resultsHandle);
   }
   @Override
-  public <T> T columnInfoAs(Plan plan, Class<T> as) {
+  public <T> T columnInfoAs(PlanBuilder.PreparePlan plan, Class<T> as) {
     ContentHandle<T> handle = handleFor(as);
     if (!(handle instanceof JSONReadHandle))
       throw new IllegalArgumentException("The handle is not an instance of JSONReadHandle.");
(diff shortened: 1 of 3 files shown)


llinggit:
The server side review: https://crucible.marklogic.com/cru/CR-878

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
```diff
@@ -285,6 +285,30 @@ public RawSPARQLSelectPlanImpl newRawSPARQLSelectPlan(TextWriteHandle handle) {
     return handle.get();
   }
 
+  @Override
+  public <T extends JSONReadHandle> T columnInfo(Plan plan, T resultsHandle) {
```
ehennum, line 289: Should the type of the first parameter be Plan or PreparePlan?

PlanBuilder.map() returns ExportablePlan. Does column info return useful information for ExportablePlan?  If so, Plan is the right type.

Otherwise, consistent with the unit test, PreparePlan is the right type.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
```diff
@@ -285,6 +285,30 @@ public RawSPARQLSelectPlanImpl newRawSPARQLSelectPlan(TextWriteHandle handle) {
...
+    AbstractWriteHandle astHandle = requestPlan.getHandle();
+
+    RequestParameters params = new RequestParameters();
+    params.add("output",     "columnInfo");
+
+    return services.postResource(requestLogger, "rows", null, params, astHandle, resultsHandle);
+  }
+  @Override
+  public <T> T columnInfoAs(Plan plan, Class<T> as) {
+    ContentHandle<T> handle = handleFor(as);
```
ehennum, line 304: Before casting to a JSONReadHandle, would it be appropriate to check JSONReadHandle.class.isInstance(handle) and throw an error if, for instance, the class is not an XMLDomHandle?

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -1205,6 +1206,27 @@ public void testMapper() throws IOException, XPathExpressionException {
...
+                                    p.cts.jsonPropertyWordQuery("instrument", "trumpet"),
+                                    p.cts.jsonPropertyWordQuery(p.xs.string("lastName"), p.xs.stringSeq("Armstrong", "Davis"))
+                            )
+                    )
+                    .orderBy(p.col("lastName"));
+
+    String result = rowMgr.columnInfo(builtPlan, new StringHandle()).get();
+    assertNotNull(result);
+    result = rowMgr.columnInfoAs(builtPlan, String.class);
+    assertNotNull(result);
```
ehennum, line 1227: Should the assertions verify that the returned information identifies all of the expected columns and thus is useful?

To put it the other way, is it possible that an invalid response could be non-null?

ehennum (changes requested):
Looks very close to me.  Please consider the comments and, after disposition, merge the changes.

Thanks for providing the capability.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
llinggit: Hi Erik, I'm a little confused here. My unit test used PlanBuilder.ExportablePlan and it works, so I think columnInfo() works with ExportablePlan.  And what is PlanBuilder.map()? Thanks.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
llinggit: Thanks for the suggestion. I'll add the check here.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
llinggit: Thanks for the comment. I'll add more checking here.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
ehennum: Hi, Ling:  Sorry, I meant PlanBuilder.PreparePlan.map​() -- the client method equivalent to protptype.map() on the enode.

In the unit test, is the actual type of the object ExportablePlan?  Or is the actual type of the object a super interface of ExportablePlan assigned to a variable of type ExportablePlan?  I ask because the actual type of the object and not the type of the assigned variable is what matters for validation.

I ask because my guess is that column info can't be returned on the plan generated by map(), given that map could return anything.  If that's correct, then the parameter shouldn't accept the plan objects constructed by PlanBuilder.PreparePlan.map​()

Issue #1342, opened 2022-04-18
Enabling compression and chunking in MarkLogic 10 in Java Client API


llinggit:
Please refer to https://bugtrack.marklogic.com/57132
"It has been decided that for ML10 we will not automatically turn on HTTP chunking/compression in ML10 based on  request headers, see; https://bugtrack.marklogic.com/57123   However compression will improve user experience and chunking allow customers to detect errors while streaming. 

A lot of customers are using CURL, Mulesoft, TIBCO, to download (and transport) datasets from ML, so we need to add an option to 'high volume' endpoints (e.g. /v1/rows) to force enable gzip/chunking. Those endpoint(s) should call: xdmp:set-response-chunked-encoding(fn:true()) and/or xdmp:set-response-compression(fn:true()) . This code is already in trunk/b10 so there is no need to wait on this feature. 

I would recommend to have an option which also work outside the ML JavaClient API (e.g.  an URL parameter), because most customers (in my experience) use standard tools for this (such as CURL, TIBCO etc) to do this plumbing. Others use indirectly the ML JavaClient API (e.g. NIFI), but then this requires a new NIFI connector "compiled" to a newer Java Client which is more difficult than adding a URL parameter. Typically those tools already support chunked/compression or they have options to support this, e.g. by feeding the input through a unzip workflow.
"

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

PR #1344, opened 2022-04-18
Enabling compression and chunking #1342

llinggit:
* What issue are you addressing with this pull request? #1342 
* Version of MarkLogic Java Client API (see Readme.txt) latest
* Version of MarkLogic Server (see admin gui on port 8001) 10.0-9.1
* Java version (`java -version`) 1.8
* OS and version Mac

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
@@ -4502,22 +4502,33 @@ public class OkHttpServices implements RESTServices {
     String operation, String entityType, Response response) {
     if ( response == null ) return null;
     ResponseBody body = response.body();
-    MimeMultipart entity = body.contentLength() != 0 ?
+    long length = body.contentLength();
+    MimeMultipart entity = length != 0 ?
       getEntity(body, MimeMultipart.class) : null;
 
+    try {
+        if (length == -1 && entity != null) entity.getCount();
+    } catch (MessagingException e) {
+        entity = null;
+    }
     List<BodyPart> partList = getPartList(entity);
-      try {
-          Headers trailer = response.trailers();
-          String code = trailer.get("ml-error-code");
-          String msg = trailer.get("ml-error-message");
-          String sha = trailer.get("ml-content-sha256");
 
-          if (code != null && !"N/A".equals(code)) {
-              throw new RuntimeException("code = " + code + ", msg = " + msg + ", sha256 = " + sha);
-          }
-      } catch (IOException e) {
-          e.printStackTrace();
-      }
+    try {
+        Headers trailer = response.trailers();
+        String code = trailer.get("ml-error-code");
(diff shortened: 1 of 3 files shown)


marklogic-client-api/src/test/resources/bootstrap.xqy
```diff
@@ -578,6 +578,8 @@ declare function bootstrap:security-config() {
         ) else (
             if ($user eq "valid") then (
                 bootstrap:security-eval('sec:create-user("valid", "valid unprivileged user", "x", (), (), (), ())')
+            ) else if ($user eq "rest-reader") then (
+                bootstrap:security-eval('sec:create-user("rest-reader", "rest-reader user", "x", ("rest-reader", "rest-reader-internal"), (), (), ())')
```
ehennum, line 582: In a customer adoption, an external user should never need a *-internal role.  

Instead, the security model is that code on the server should amp to the *-internal role for users that have the external role.

If the user for tests has the *-internal role, then the tests don't reliably indicate whether production users can execute the requests.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -465,6 +465,24 @@ public void testView() {
...
   }
+
+  @Test
+  public void testSQLException() {
+    RowManager rowMgr = Common.client.newRowManager();
+
+    PlanBuilder p = rowMgr.newPlanBuilder();
+
+    PlanBuilder.ExportablePlan builtPlan =
+            p.fromSql("select case when lastName = 'Davis' then fn_error(fn_qname('ERROR','test')) end, opticUnitTest.musician.* from (select * from opticUnitTest.musician order by lastName)");
```
ehennum, line 476: Thanks for providing this missing unit test.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -4501,6 +4506,18 @@ private <U extends OkHttpResultIterator> U makeResults(
...
 
     List<BodyPart> partList = getPartList(entity);
+      try {
+          Headers trailer = response.trailers();
+          String code = trailer.get("ml-error-code");
+          String msg = trailer.get("ml-error-message");
+          String sha = trailer.get("ml-content-sha256");
+
+          if (code != null && !"N/A".equals(code)) {
+              throw new RuntimeException("code = " + code + ", msg = " + msg + ", sha256 = " + sha);
```
ehennum, line 4516: Is it possible to map to one of the MarkLogic Java exceptions in the same cases that would have this mapping without a chunked response?

That is, is it possible to refactor the existing exception construction code to be callable with either an HTTP error code and message or a trailer error code and message?

ehennum (approved):
In general, the approach looks good.  I've raised a couple of concerns about the security model and about the Java exceptions.

Out of curiosity, what was the resolution on the apparent multipart parsing issue?

PR #1354, opened 2022-09-13
JAVA-232 : FromDocUris implementation in java-client-api

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-232 
* Have you run unit tests?  Yes
* Version of MarkLogic Java Client API (see Readme.txt) - 5.3.3 
* Version of MarkLogic Server (see admin gui on port 8001) - ML 11.0
* Java version (`java -version`) - Java 1.8
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
@@ -151,7 +151,15 @@ public interface PlanBuilderBase {
      * @return  an AccessPlan object
      */
     PlanBuilder.AccessPlan fromLiterals(Map<String,Object>[] rows, XsStringVal qualifierName);
-
+    /**
+     * Provides a convenience equivalent that is equivalent to chaining
+     * {@link PlanBuilder.AccessPlan#fromLexicons(Map, String)},
+     * and {@link PlanBuilder.ModifyPlan#where(CtsQueryExpr)}.
+     * @param querydef This parameter is a cts query.
+     * @param qualifierName  Specifies a name for qualifying the column names similar to a view name.
+     * @return  an AccessPlan object
+     */
+    PlanBuilder.AccessPlan fromDocUris(CtsQueryExpr querydef, String qualifierName);
     /**
      * This function constructs a JSON object with the specified properties. The object can be used as the value of a column in a row or passed to a builtin function.
      * <p>
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
@@ -163,6 +163,14 @@ public class PlanBuilderSubImpl extends PlanBuilderImpl {
   }
 
   @Override
+  public AccessPlan fromDocUris(CtsQueryExpr querydef, String qualifierName) {
+    return new AccessPlanSubImpl(
+            this, "op", "from-doc-uris", new Object[]{querydef, (qualifierName == null) ?
+            null : xs.string(qualifierName)}
+    );
+  }
+
+  @Override
   public PlanAggregateCol avg(String name, String column, PlanValueOption option) {
     return avg(col(name), col(column), option);
   }
@@ -918,6 +926,7 @@ public class PlanBuilderSubImpl extends PlanBuilderImpl {
         case "from-search":
         case "from-search-docs":
         case "from-triples":
+        case "from-doc-uris":
           if (fnArgs.length < 1) {
             throw new IllegalArgumentException("accessor constructor without parameters: "+fnArgs.length);
           }
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
@@ -39,7 +39,12 @@ import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.xpath.XPathExpressionException;
 
+import com.marklogic.client.datamovement.DataMovementManager;
+import com.marklogic.client.datamovement.WriteBatcher;
+import com.marklogic.client.io.*;
 import com.marklogic.client.io.marker.JSONReadHandle;
+import com.marklogic.client.query.DeleteQueryDefinition;
+import com.marklogic.client.query.QueryManager;
 import com.marklogic.client.row.*;
 import com.marklogic.client.type.*;
 import org.junit.AfterClass;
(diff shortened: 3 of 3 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -791,4 +791,6 @@ interface ModifyPlanBase {
      */
     interface PreparePlanBase {
     }
+
+  public abstract PlanBuilder.AccessPlan fromDocUris(CtsQueryExpr querydef, String qualifier);
```
rjrudin, line 795: I think this should be with the other "from" methods at the top of `PlanBuilderBase` - I'd put it after `fromLiterals(Map)` at line 153 and before `jsonObject`. 

Also, we need javadoc comments for this, correct? I'm wondering what `qualifier` in particular is - if it's the same as `qualifierName` in the other `from` methods, then it should be renamed to `qualifierName` for consistency.

marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
```diff
@@ -482,6 +482,13 @@ public ServerExpression seq(ServerExpression... expr) {
     return new BaseTypeImpl.ServerExpressionListImpl(expr, true);
   }
 
+  @Override
```
rjrudin, line 485: Same thing as in the interface - I think this should be grouped with the other `from` methods, so I'd put this at line 165 after `fromView` and before `avg`.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -200,6 +200,12 @@ public static void beforeClass() throws IOException, InterruptedException {
   }
   @AfterClass
   public static void afterClass() {
+    DocumentManager docMgr = Common.client.newDocumentManager();
```
rjrudin, line 203: I've found the best practice for tests is for data to be deleted before a test begins and not deleted after a test ends. Leaving the test data in place makes debugging a lot easier when the test fails. And it's often useful to run a test to generate data and then fiddle with it in qconsole. 

I know it can be difficult to switch to that practice when an existing test suite doesn't already delete data at the start of the tests. But given that this test class was already creating data and not deleting it (which I think is good), I would test to see if it's fine to leave these documents around. If they break some other test, then that test should just clear out the database before it begins.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -1479,6 +1485,57 @@ public void testBucketGroup() {
...
+    PlanBuilder.ExportablePlan builtPlan = p.fromDocUris(p.cts.wordQuery("trumpet"), "");
+
+    RowSet<RowRecord> recordRowSet = rowMgr.resultRows(builtPlan);
+    Set<String> uriSet = new HashSet<>();
+    uriSet.add("/optic/test/musician4.json");
+    uriSet.add("/optic/test/musician1.json");
+    Iterator<RowRecord> rows = rowMgr.resultRows(builtPlan).iterator();
+    while (rows.hasNext()){
+      String temp = rows.next().get("uri").toString().replaceAll("^\"|\"$", "");
+      Iterator<String> itr = uriSet.iterator();
```
rjrudin, line 1502: I think this line can be removed.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -1479,6 +1485,57 @@ public void testBucketGroup() {
...
+    PlanBuilder p = rowMgr.newPlanBuilder();
+    PlanBuilder.ExportablePlan builtPlan = p.fromDocUris(p.cts.wordQuery("trumpet"), "");
+
+    RowSet<RowRecord> recordRowSet = rowMgr.resultRows(builtPlan);
+    Set<String> uriSet = new HashSet<>();
+    uriSet.add("/optic/test/musician4.json");
+    uriSet.add("/optic/test/musician1.json");
+    Iterator<RowRecord> rows = rowMgr.resultRows(builtPlan).iterator();
+    while (rows.hasNext()){
+      String temp = rows.next().get("uri").toString().replaceAll("^\"|\"$", "");
```
rjrudin, line 1501: What's the replaceAll needed for?

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -1479,6 +1485,57 @@ public void testBucketGroup() {
...
+
+  @Test
+  public void testFromDocUrisWithDirectoryQuery() throws Exception {
+    RowManager rowMgr = Common.client.newRowManager();
+    PlanBuilder p = rowMgr.newPlanBuilder();
+    PlanBuilder.ExportablePlan builtPlan = p.fromDocUris(p.cts.directoryQuery("/testFromDocUrisWithDirectoryQuery/"), "");
+    DocumentMetadataHandle meta = new DocumentMetadataHandle().withCollections("testcollection");
+    DataMovementManager moveMgr = Common.client.newDataMovementManager();
+    WriteBatcher batcher = moveMgr.newWriteBatcher();
+    batcher.addAs("/testFromDocUrisWithDirectoryQuery/doc1.txt", meta,"This is doc1");
```
rjrudin, line 1516: I think this test would be a little better if at least one of the 4 docs does not belong to the directory that we're going to query on.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
anu3990: replaceAll will get rid of the double quotes and the variable temp will be - /optic/test/musician4.json instead of "/optic/test/musician4.json". Without the replaceAll the asserts will fail.

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -151,7 +151,13 @@ public interface PlanBuilderBase {
      * @return  an AccessPlan object
      */
     PlanBuilder.AccessPlan fromLiterals(Map<String,Object>[] rows, XsStringVal qualifierName);
-
+    /**
+     * Provides a convinience equivalent to fromLexicons({uri: cts.uriReference()},qualifier,op.fragmentIdCol('fragmentId')).where(querydef).
```
rjrudin, line 155: Typo - convinience -> convenience 

Also, for consistency with other javadocs in this class, I think "link" should be used - i.e.:

    equivalent to {@link PlanBuilder#fromLexicons(Map, String, PlanSystemColumn)}

And we'd want to reference that Java method as opposed to an SJS script, as the user may not be familiar at all with SJS.

rjrudin (approved):

PR #1355, opened 2022-09-14
Fixing graph tests

rjrudin:
Resolves JAVA-225. Modified the bootstrap script to create a `rest-delete-graph` role that grants the `term-query`privilege to the `rest-writer` user, thus allowing that user to delete graphs. 

Few other changes in here:

- Replaced some hardcoded instances of "localhost" and "admin" with the variables in Common (I was testing this on an AWS instance and thus localhost/admin)
- Removed unused `changeAssignmentPolicy` method in `ConcurrencyTest` (I noticed the hardcoded password in there, but then deleted the whole method since it's unused)
- Fixed RowManagerTest, as the SQL statement wasn't causing an error, so changed it to something that will cause an error


rjrudin:
@anu3990 I tested this on 10.0-8.3 and 10.0-9.2; tests like `GraphsTest` were failing on both of those versions of ML because the `rest-writer` user did not have the `term-query` privilege. I just installed ML 10.0-9.4 and the tests are not failing (although `RowManagerTest` still fails, as it is on Jenkins, and this PR has a fix for that). 

I am thinking we should still include this fix because I can't find anything in bugtrack indicating that this fix occurred. I am going to add a comment to where the rest-delete-graph role is created to explain why it's being created.

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
```diff
@@ -523,7 +523,7 @@ public void testSQLException() {
     RowManager rowMgr = Common.client.newRowManager();
     PlanBuilder p = rowMgr.newPlanBuilder();
     PlanBuilder.ExportablePlan builtPlan =
-        p.fromSql("select case when lastName = 'Davis' then fn_error(fn_qname('', 'SQL-TABLENOTFOUND'), 'Internal Server Error') end, opticUnitTest.musician.* from (select * from opticUnitTest.musician order by lastName)");
```
llinggit, line 526: I just checked the Jenkins job, this test passes in 11, but failed in 10. It should be excluded from running in ML 10. We should add an annotation as you mentioned, but not modify it I think.

llinggit (approved):
Just that RowManagerTest.java. Otherwise it looks good to me.

anu3990 (approved):

marklogic-client-api/src/test/java/com/marklogic/client/test/RowManagerTest.java
rjrudin: Thanks @llinggit - given that it's 1 of 3 tests passing on 11 but not on 10, I'm going to remove this change completely and tackle all 3 tests in a separate JIRA / PR.

PR #1356, opened 2022-09-16
Converted bootstrap program into ml-gradle application

rjrudin:
Provides the following benefits:

- No longer depends on localhost and admin/admin for the host and admin user; can override these via gradle-local.properties
- Can redeploy the app after making changes to any part of the test app config
- Can undeploy the app

Data documents created by the former bootstrap.xqy module are now found under src/test/ml-data.


BillFarber (approved):

marklogic-client-api/src/main/resources/Example.properties
```diff
@@ -3,8 +3,8 @@ example.reader_user=rest-reader
 example.reader_password=x
 example.writer_user=rest-writer
 example.writer_password=x
-example.admin_user=rest-admin
-example.admin_password=x
+example.admin_user=admin
```
llinggit, line 6: Have you tried running the example tests with user admin? I'm not sure about this part. At least I know for some unit tests, admin doesn't have the permission to read the data loaded. Not sure about the example tests though.

llinggit (approved):
Only has a concern about the user change in example tests. If all tests pass, then that's fine.

marklogic-client-api/src/main/resources/Example.properties
rjrudin: This is only used by the example programs, and at least the tests for those examples are all passing. The main concern with the "admin" user is that it can't read data via a TDE template unless it has at least one role matching the permissions on documents exposed via the TDE template. That does not yet seem to be an issue for the example programs.

If that does become an issue, I think the solution would be to introduce "rest_admin_user" and "rest_admin_password" as I think "admin_user" and "admin_password" are universally understood to refer to an "admin" user, not a "rest_admin" user.

PR #1359, opened 2022-09-20
DEVO-635: Setting up pipeline to run unit tests on javaclient api

SameeraPriyathamTadikonda:


Jenkinsfile
```diff
@@ -0,0 +1,37 @@
...
+          export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
+          cd java-client-api
+          ./gradlew mlDeploy
+        '''
+        sh label:'test', script: '''#!/bin/bash
+          export JAVA_HOME=$JAVA_HOME_DIR
+          export GRADLE_USER_HOME=$WORKSPACE/$GRADLE_DIR
+          export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
+          cd java-client-api
+          ./gradlew test  || true
```
rjrudin, line 31: @SameeraPriyathamTadikonda I think this should just be marklogic-client-api:test - I don't think we need to run the tests for the other two projects yet, but maybe later.

rjrudin (changes requested):

Jenkinsfile
SameeraPriyathamTadikonda: I changed to run functional tests to run after the PR merges.

Jenkinsfile
```diff
@@ -0,0 +1,53 @@
...
+  environment{
+    JAVA_HOME_DIR="/home/builder/java/openjdk-1.8.0-262"
+    GRADLE_DIR   =".gradle"
+    DMC_USER     = credentials('MLBUILD_USER')
+    DMC_PASSWORD = credentials('MLBUILD_PASSWORD')
+  }
+  stages{
+    stage('tests'){
+      steps{
+        copyRPM 'Release','10.0-9.4'
```
rjrudin, line 17: @anu3990  @llinggit  Do you think we should use the latest ML release or the latest nightly build?

llinggit: sorry I missed the context here. We've already got a Jenkins job to run java unit tests against nightly build, right?

rjrudin: This will run against submitted PRs before they're merged

SameeraPriyathamTadikonda: We can also run the tests on a nightly server after merging the PR to develop.

llinggit: nightly build is better I think.

rjrudin (approved):
LGTM, and I'm good with it going against the nightly instead

PR #1363, opened 2022-09-23
JAVA-246: Initial support for fromParam

rjrudin:
Have only tested binding an array of JSON documents, need to cover a lot more scenarios. But wanted to get some initial feedback on the implementation approach.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
@@ -17,6 +17,9 @@ package com.marklogic.client.expression;
 
 import java.util.Map;
 
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.marklogic.client.document.DocumentWriteSet;
+import com.marklogic.client.io.marker.AbstractWriteHandle;
 import com.marklogic.client.io.marker.JSONReadHandle;
 
 import com.marklogic.client.type.*;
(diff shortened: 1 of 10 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -558,6 +590,8 @@ interface PlanBase {
      * Use AccessPlan as the type for instances of AccessPlan.
      */
     interface AccessPlanBase {
+        PlanBuilder.Plan bindParam(String param, AbstractWriteHandle content);
+        Map<String, AbstractWriteHandle> getContentParams();
```
rjrudin, line 594: This is the mechanism for allowing a user to bind a param containing what I am calling "content", and I think an AbstractWriteHandle is the right choice here, as that gives the user plenty of flexibility without having to construct a DocumentWriteSet.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -3957,6 +3957,51 @@ public RESTServiceResultIterator postIteratedResource(RequestLogger reqlog,
       reqlog, path, transaction, params, input, outputMimetypes);
   }
 
+    public RESTServiceResultIterator postMultipartForm(
```
rjrudin, line 3960: This copies a good chunk of `postIteratedResourceImpl`, but I didn't want to factor out the common stuff until I know this is working for all scenarios. The main thing this does is construct a multipart/form-data request.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -3957,6 +3957,51 @@ public RESTServiceResultIterator postIteratedResource(RequestLogger reqlog,
...
+      if ("rows".equals(path)) {
+          requestBldr.addHeader("ML-Check-ML11-Headers", "true");
+          requestBldr.addHeader("TE", "trailers");
+      }
+
+      Function<Request.Builder, Response> doPostFunction = funcBuilder ->
+              doPost(reqlog, funcBuilder.header(HEADER_ACCEPT, multipartMixedWithBoundary()), multiBuilder.build());
+
+      // The construction of this was based on whether the primary input was resendable or not. We don't have a primary
+      // input with a multipart request. So keeping this as null for now.
```
rjrudin, line 3996: Wasn't sure what to do here.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -5645,6 +5692,21 @@ public void writeTo(BufferedSink sink) throws IOException {
     }
   }
 
+  public static RequestBody makeRequestBodyForContent(AbstractWriteHandle content) {
```
rjrudin, line 5695: Extracted this from the Bulk DS code so it can be used for building a multipart/form-data request.

marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
```diff
@@ -377,7 +378,13 @@ private RESTServiceResultIterator makeRequest(
     addDatatypeStyleParam(params,     datatypeStyle);
     addRowStructureStyleParam(params, rowStructureStyle);
 
-// QUESTION: outputMimetypes a noop?
+    if (requestPlan instanceof PlanBuilderBase.AccessPlanBase) {
+      Map<String, AbstractWriteHandle> contentParams = ((PlanBuilderBase.AccessPlanBase)requestPlan).getContentParams();
```
rjrudin, line 382: This seems like the right place to insert the multipart branch - i.e. if the plan doesn't have any content params, then we can do a regular post as there's no need for multipart. But if there's at least one content param, we know we need a multipart request, so we go through a different branch of code.

anu3990 (approved):

rjrudin:
@anu3990  I'll merge this in, but the feature will definitely change as I add support for attachments.

PR #1369, opened 2022-09-28
JAVA-246: RowBatcher and QueryBatcher now provide access to timestamp

rjrudin:
Allows for client to perform subsequent operations using that same timestamp.

Note that QueryBatcher provides access to this via the argument passed to QueryBatchListener, though that of course requires a bit more effort for a user to get than simply calling getServerTimestamp after the job completes.


rjrudin:
I pulled this into 5.6.0 in preparation for needing it for our future query capability for Kafka and any other connector that will use `RowBatcher`. This simple getter is critical for allowing the connector to follow up with a query for the latest dateTime value in a user-defined index so that the on the subsequent job run, the connector can constrain the user's query to only those records with a dateTime greater than the looked-up value from the current job run.

BillFarber (approved):

anu3990 (approved):

rjrudin:
Ack - this used the wrong JIRA ID, should be JAVA-234

PR #1371, opened 2022-09-30
JAVA-246: Moving new bindParam methods to Plan

rjrudin:
Had to implement the methods in RawPlanImpl since it does not have a common parent with PlanSubImpl. Added a test for a raw plan to cover this new code.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
@@ -47,7 +47,6 @@ import com.marklogic.client.MarkLogicInternalException;
 import com.marklogic.client.Transaction;
 import com.marklogic.client.expression.PlanBuilder;
 import com.marklogic.client.expression.PlanBuilder.Plan;
-import com.marklogic.client.expression.PlanBuilderBase;
 import com.marklogic.client.impl.RESTServices.RESTServiceResult;
 import com.marklogic.client.impl.RESTServices.RESTServiceResultIterator;
 import com.marklogic.client.io.*;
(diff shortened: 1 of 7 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/RowManagerImpl.java
```diff
@@ -1718,6 +1718,26 @@ public Plan bindParam(PlanParamExpr param, PlanParamBindingVal literal) {
 
       return parameterize(getHandle(), nextParams);
     }
+
+    @Override
+    public Plan bindParam(String param, AbstractWriteHandle content) {
+      throw new UnsupportedOperationException("TODO Unclear if this needs to be supported for a raw plan");
```
rjrudin, line 1724: @anu3990 @ehennum @llinggit  I don't know if `RawPlanImpl` should support these methods or not. Does it make sense for a raw plan to have content and attachments bound to it? I don't know the answer.

ehennum: Yes, @rjrudin, that's the reason the specification of the structure of the input data in fromParam() and of the subsequent processing of the input data in the rest of the plan is separate from the provision of the input data in result() or execute().

The same plan can be used to process any number of batches of data.

In the future, after RowManager supports fromParam(), RowBatcher could be constructed with the plan but take a stream of rows to be batched and sent to the server.

Makes sense?

rjrudin: Thanks @ehennum - does that mean that these methods should be implemented in `RawPlanImpl`, basically in the same fashion as they are in `PlanSubImpl`? 

By the way, I'm testing out a `bindParam(param, DocumentWriteSet)` right now - that should work fine, there are just a couple questions with it that we can go over today.

ehennum: Apt point, @rjrudin .  Yes, I would think that PlanBase should declare the new bindParam() methods such that PlanBuilder.Plan and thus RawPlan inherits them (and thus RawPlanImpl would implement them as in PlanSubImpl, ideally through shared implementation functions and classes).

anu3990 (approved):

PR #1382, opened 2022-10-12
Ignoring a couple Optic export tests

rjrudin:
Also deleted obsolete Eclipse files, which shouldn't have been added to git (they were giving VSCode a real headache).


rjrudin:
Ignoring all of these until the export bug is fixed, will re-enable again soon

PR #1392, opened 2022-10-17
Updating tests based on 57902 fix

rjrudin:
When using `write()`, rows can now be returned.


rjrudin:
Merging this so I can get a clean run on #1390

Issue #1405, opened 2022-10-24
WriteBatcher does not write documents when total documents provided is less than 100 and without using withBatchSize


anu3990:
So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

### Input: Some code to illustrate the problem, preferably in a state that can be independently reproduced on our end

DataMovementManager moveMgr = Common.client.newDataMovementManager();
        WriteBatcher writeBatcher = moveMgr.newWriteBatcher();
        DocumentMetadataHandle meta = new DocumentMetadataHandle().withCollections("RowManagerValidateDocTest");
        moveMgr.startJob(writeBatcher);
        for(int i=0; i<99; i++){
            writeBatcher.addAs("test"+i, meta, new StringHandle("<Doc><key>"+i+"</key><Value>value"+i+"</Value></Doc>").withFormat(XML));
        }
        writeBatcher.awaitCompletion();
        moveMgr.stopJob(writeBatcher);

### Actual output: What did you observe? What errors did you see? Can you attach the logs? (Java logs, MarkLogic logs)
No documents were written to the database

### Alternatives: What else have you tried, actual/expected? 
If the total number of documents was >= 100, all the documents were written. Also, when we use withhBatchSize, all the documents were written. For example, all the 99 documents were written to the database -

DataMovementManager moveMgr = Common.client.newDataMovementManager();
        WriteBatcher writeBatcher = moveMgr.newWriteBatcher().withBatchSize(99);
        DocumentMetadataHandle meta = new DocumentMetadataHandle().withCollections("RowManagerValidateDocTest");
        moveMgr.startJob(writeBatcher);
        for(int i=0; i<99; i++){
            writeBatcher.addAs(DIRECTORY+i, meta, new StringHandle("<Doc><key>"+i+"</key><Value>value"+i+"</Value></Doc>").withFormat(XML));
        }
        writeBatcher.awaitCompletion();
        moveMgr.stopJob(writeBatcher);

anu3990:
We can use flushAndWait instead of awaitCompletion.

PR #1406, opened 2022-10-24
WriteBatcher does not write documents when total documents provided is less than 100 and without withBatchSize

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1405
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/BatcherImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/BatcherImpl.java
@@ -24,7 +24,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public abstract class BatcherImpl implements Batcher {
   private String jobName = "unnamed";
   private String jobId = null;
-  private int batchSize = 100;
+  private int batchSize = 0;
   private int threadCount = 1;
   private ForestConfiguration forestConfig;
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/datamovement/impl/BatcherImpl.java
```diff
@@ -24,7 +24,7 @@
 public abstract class BatcherImpl implements Batcher {
   private String jobName = "unnamed";
   private String jobId = null;
-  private int batchSize = 100;
+  private int batchSize = 0;
```
rjrudin, line 27: I don't think this change can be made at this level; a QueryBatcherTest test is failing due to a divide-by-zero change. At minimum, I think this change needs to happen within WriteBatcherImpl so that it does not impact RowBatcher or QueryBatcher. 

Also, there's logging indicating that this is immediately set to a batch size of 1- "11:00:52.984 [main] WARN  c.m.c.d.impl.WriteBatcherImpl - batchSize should be 1 or greater--setting batchSize to 1". We wouldn't want that appearing any time the user doesn't explicitly set batch size. 

I'll ping you internally about this, as I'm not sure that the current behavior is a problem - i.e. the test you wrote can easily be made to pass by using `flushAndWait` instead of `awaitCompletion`.

rjrudin (changes requested):

PR #1409, opened 2022-10-25
JAVA-242 : Implement "validateDoc" pipeline operation

anu3990:
* What issue are you addressing with this pull request? https://project.marklogic.com/jira/browse/JAVA-242 
* Are you modifying the correct branch? yes "develop"
* Version of MarkLogic Server (see admin gui on port 8001) 11.0
* Java version (`java -version`)- java 1.8
* OS and version

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
@@ -465,6 +465,18 @@ public interface PlanBuilderBase {
     TransformDef transformDef(String path);
 
     /**
+     * Build a schema definition for use with {@code validateDoc}.
+     *
+     * @return a new {@code SchemaDefExpr}
+     */
+    SchemaDefExpr schemaDefinition(String kind);
+    /**
+     * Build a error definition for use with {@code validateDoc}.
+     *
+     * @return a new {@code PlanErrorDisposition}
+     */
+    PlanErrorDisposition errorDispositionDef(int logSize, String logLevel);
+    /**
      * Convenience method for constructing a permission that can then be used e.g. with {@code jsonArray} when binding
      * an array of permissions to a column.
      *
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/SchemaDefImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/SchemaDefImpl.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2022 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.client.impl;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.expression.SchemaDefExpr;
+
+public class SchemaDefImpl implements SchemaDefExpr, BaseTypeImpl.BaseArgImpl {
+    private String kind;
+    private String mode;
+    private String schemaUri;
+
+    public SchemaDefImpl(String kind){
+        this.kind = kind;
+    }
+    @Override
+    public StringBuilder exportAst(StringBuilder strb) {
+        ObjectNode node = new ObjectMapper().createObjectNode();
+        if(this.kind != null)
+            node.put("kind", this.kind);
+        if(this.mode != null)
+            node.put("mode", this.mode);
+        if(this.schemaUri != null)
+            node.put("schemaUri", this.schemaUri);
+        return strb.append(node.toString());
+    }
+
+    @Override
+    public SchemaDefExpr withMode(String mode) {
+        this.mode = mode;
+        return this;
+    }
+
+    @Override
+    public SchemaDefExpr withSchemaUri(String schemaUri) {
+        this.schemaUri = schemaUri;
+        return this;
+    }
+}
--- a/marklogic-client-api/src/main/java/com/marklogic/client/expression/SchemaDefExpr.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/expression/SchemaDefExpr.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2022 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.client.expression;
+
+import com.marklogic.client.type.PlanSchemaDef;
+
+/**
+ * Defines the schema for using with the {@code validateDoc} operator;
+ */
+public interface SchemaDefExpr extends PlanSchemaDef {
+    /**
+     * Defines the mode property.
+     *
+     * @param mode -
+     *      xmlSchema - the mode property takes a strict, lax or type value.
+     *      jsonSchema or schematron - the mode property takes a full or strict value
+     * @return - an instance of SchemaDefExpr
+     */
+    SchemaDefExpr withMode(String mode);
+    /**
+     * Defines the schema based on an existing document in the database.
+     *
+     * @param schemaUri - property takes a string with the URI of the schema
+     * @return - an instance of SchemaDefExpr
+     */
+    SchemaDefExpr withSchemaUri(String schemaUri);
+}
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/ErrorDispositionDefImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/ErrorDispositionDefImpl.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2022 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.client.impl;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.type.PlanErrorDisposition;
+
+public class ErrorDispositionDefImpl implements PlanErrorDisposition, BaseTypeImpl.BaseArgImpl {
+    private String logLevel;
+    private int logSize;
+
+    public ErrorDispositionDefImpl(int logSize, String logLevel) {
+        this.logLevel = logLevel;
+        this.logSize = logSize;
+    }
+
+    @Override
+    public StringBuilder exportAst(StringBuilder strb) {
+        ObjectNode node = new ObjectMapper().createObjectNode();
+        if(logLevel != null)
+            node.put("logLevel", this.logLevel);
+        node.put("logSize", this.logSize);
+        return strb.append(node.toString());
+    }
+}
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/PlanBuilderSubImpl.java
@@ -22,9 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.document.DocumentWriteOperation;
 import com.marklogic.client.document.DocumentWriteSet;
-import com.marklogic.client.expression.PlanBuilder;
-import com.marklogic.client.expression.SemExpr;
-import com.marklogic.client.expression.TransformDef;
+import com.marklogic.client.expression.*;
 import com.marklogic.client.io.marker.AbstractWriteHandle;
 import com.marklogic.client.io.marker.ContentHandle;
 import com.marklogic.client.io.marker.JSONReadHandle;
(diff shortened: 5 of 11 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/expression/ValidateDocErrorDispositionDef.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Defines an errorDisposition for using with the {@code validateDoc} operator;
+ */
+public interface ValidateDocErrorDispositionDef extends PlanErrorDisposition {
+    /**
+     * Define the size of the logs.
+     *
+     * @param logSize - The value may be a non-negative JavaScript integer from 0 (the default, which suppresses logging)
+     *    to Number.MAX_SAFE_INTEGER (9007199254740991; effectively summarizing all errors in a single entry).
```
rjrudin, line 28: I know this is copied from the functional spec, but as a user, I don't understand what "size" means in this context. Does it refer to the number of errors logged, or the physical size of each log message?

marklogic-client-api/src/main/java/com/marklogic/client/expression/ValidateDocErrorDispositionDef.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return - an instance of ValidateDocErrorDispositionDef
+     */
+    ValidateDocErrorDispositionDef withLogSize(int logSize);
+    /**
+     * Define the property of the error disposition object
+     *
+     * @param logLevel - specifies one of the following:
+     *   summary – log only the document uri and error message for each error
```
rjrudin, line 36: Should use "URI" instead of "uri" in any docs.

marklogic-client-api/src/main/java/com/marklogic/client/expression/ValidateDocSchemaDefinition.java
```diff
@@ -0,0 +1,54 @@
...
+     * Defines the mode property.
+     *
+     * @param mode - takes a strict, lax or type value
+     * @return - an instance of ValidateDocSchemaDefinition
+     */
+    ValidateDocSchemaDefinition withMode(String mode);
+    /**
+     * Defines the schema based on an existing document in the database.
+     *
+     * @param schemaUri - property takes a string with a schematron URI
```
rjrudin, line 50: If this is specific to a Schematron URI, then this should be called `withSchematronURI`. If it's generic to any schema, then the name is good and the comment should say "with a schema URI".

marklogic-client-api/src/main/java/com/marklogic/client/expression/ValidateDocSchemaDefinition.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    ValidateDocSchemaDefinition withKind(String kind);
+    /**
+     * Defines the schema as a node object.
+     *
+     * @param schema - key value pair defining a schema
+     * @return - an instance of ValidateDocSchemaDefinition
+     */
+    // TODO : add tests when https://bugtrack.marklogic.com/58025 is fixed
+    ValidateDocSchemaDefinition withSchema(Map<Object, Object> schema);
```
rjrudin, line 39: Is the expectation that the user would define a JSON object? If so, I think we should accept an `ObjectNode` here which is much easier for a user to construct.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+
+import static com.marklogic.client.io.Format.JSON;
+import static com.marklogic.client.io.Format.XML;
+import static org.junit.Assert.assertTrue;
+
+public class RowManagerValidateDocTest {
```
rjrudin, line 46: Have this extend `AbstractOpticUpdateTest`, which simplifies a number of things, including deleting all docs in the "/acme/" directory before a test starts. Then, if you have to create any docs, use "/acme/" as a prefix in your URI so that you don't have to worry about any such doc existing when your test starts. 

Also, for consistency with the other tests, just call it "ValidateDocTest".

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        // TODO: uncomment the below after https://bugtrack.marklogic.com/57987 is fixed
+        // assertTrue(set.size() == 5);
+
+    }
+
+    @After
+    public void cleanup(){
+        DocumentManager docMgr = Common.client.newDocumentManager();
+        docMgr.delete("/schema/jsonValidation.json", "/schematron.sch");
```
rjrudin, line 206: Instead of having to create/delete these, just add them to src/test/ml-schemas - they'll get loaded as part of `mlDeploy`. Just ensure that they'll only apply to docs created in your test.

rjrudin (changes requested):

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,212 @@
...
+    public static void setUp(){
+        Common.connect();
+        dataMovementManager = Common.client.newDataMovementManager();
+        rowManager = Common.client.newRowManager();
+        op = rowManager.newPlanBuilder();
+    }
+
+
+    @Test
+    public void validateDocWithxmlSchema() {
```
rjrudin, line 65: For each type of schema validation, we need both a positive and negative test to ensure that the validation is working correctly. Additionally, we need a separate test for each "mode" to ensure that it's working correctly as well.

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -464,6 +464,18 @@ public interface PlanBuilderBase {
      */
     TransformDefinition transformDefinition(String path);
 
+    /**
+     * Build a schema definition for use with {@code validateDoc}.
+     *
+     * @return a new {@code SchemaDefExpr}
+     */
+    SchemaDefExpr schemaDefinition(String kind);
```
rjrudin, line 472: I just renamed TransformDefinition/transformDefinition to TransformDef/transformDef to be consistent with the other types in the "client.expression" package. So I think we should do SchemaDef/schemaDef and then errorDispositionDef is good to go.

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -464,6 +464,18 @@ public interface PlanBuilderBase {
...
+    /**
+     * Build a error definition for use with {@code validateDoc}.
+     *
+     * @return a new {@code PlanErrorDisposition}
```
rjrudin, line 476: Since logLevel can be null, the javadocs should have "param"'s defined that explains that logSize is required and logLevel is null. Also need to define what are valid values for each of them.

marklogic-client-api/src/main/java/com/marklogic/client/impl/SchemaDefImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.expression.SchemaDefExpr;
+
+public class SchemaDefImpl implements SchemaDefExpr, BaseTypeImpl.BaseArgImpl {
+    private String kind;
+    private String mode;
+    private String schemaUri;
+
+    public SchemaDefImpl(String kind){
```
rjrudin, line 27: I think schemaUri should be required, and "kind" seems optional. And this class can make a very educated guess to determine what "kind" should be based on schemaUri:

1. If schemaUri ends with .mjs, then kind = mjs
2. If schemaUri ends with .xsl or .xslt, then kind = xslt

The javadocs for the `schemaDef(String schemaUri)` method can make this clear - i.e. by telling the user that they don't need to call `withKind` unless their schemaUri doesn't end with mjs, xsl, or xslt. That provides the best UX for the user, such that they only need to call `schemaDef(String schemaUri)` and then they can optionally call `withMode` as well.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.springframework.util.StringUtils;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static com.marklogic.client.io.Format.JSON;
+import static com.marklogic.client.io.Format.XML;
+import static org.junit.Assert.*;
+
+public class RowManagerValidateDocTest extends AbstractOpticUpdateTest {
```
rjrudin, line 43: For consistency with the other tests, name this "ValidateDocTest".

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,189 @@
...
+    private DataMovementManager dataMovementManager;
+
+    @Before
+    public void moreSetup(){
+        dataMovementManager = Common.client.newDataMovementManager();
+    }
+// TODO : test with rest-writer user when https://bugtrack.marklogic.com/57978 is fixed
+    @Test
+    public void xmlSchema() {
+        expectedUris.clear();
```
rjrudin, line 55: This line can be removed from each test, as expectedUris will be instantiated each time a test is run.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,189 @@
...
+        DocumentMetadataHandle meta = newDefaultMetadata();
+        dataMovementManager.startJob(writeBatcher);
+        final int uriCountToWrite = 10;
+        for (int i = 0; i < uriCountToWrite; i++) {
+            String uri = "/acme/" + i + ".xml";
+            writeBatcher.addAs(uri, meta, new StringHandle("<Doc><key>" + i + "</key><Value>value" + i + "</Value></Doc>").withFormat(XML));
+            expectedUris.add(uri);
+        }
+        writeBatcher.flushAndWait();
+        dataMovementManager.stopJob(writeBatcher);
```
rjrudin, line 66: This test would be simpler and more realistic if it instead wrote only 2 documents, one that is valid one that is invalid. That is sufficient for verifying that the schema is being used correctly to validate each document before it is written. And it avoids the need to use DMSDK to insert N documents first.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
```diff
@@ -0,0 +1,189 @@
...
+                op.schemaDefinition("xmlSchema").withMode("lax"),
+                op.errorDispositionDef(100, "summary")
+            );
+        XMLDocumentManager mgr = Common.client.newXMLDocumentManager();
+        expectedUris.forEach(uri -> assertNotNull("URI was not written: " + uri, mgr.exists(uri)));
+
+        List<RowRecord> rows = resultRows(plan);
+
+        //TODO : uncomment lines 84 and 89 after https://bugtrack.marklogic.com/57987 is fixed
+        List<String> persistedUris = rows.stream().map(row -> {
```
rjrudin, line 81: Note that this won't be an issue in the test if you're only writing 2 docs, as you'll get only 1 row back (for the valid doc) and can safely verify its contents. Though, it's more important to verify that the doc was written.

marklogic-client-api/src/main/java/com/marklogic/client/impl/SchemaDefImpl.java
rjrudin: Oops - I was thinking transformDoc! But the same logic applies, with checking for common extensions of json/xsd/sch.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
rjrudin: Also, I think this will be simpler if you just create small thesaurus documents, as shown in the docs at https://docs.marklogic.com/xdmp.validate . That way, you don't need to define an XML schema. And make the docs as small as possible - you just need 1 valid one and 1 invalid one.

rjrudin (changes requested):

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
anu3990: Hi Rob, that does not look right to me since all other files in expression package end with "Expr". Do you want to rename transformDef to TransformDefExpr?

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
anu3990: i tried without loglevel and it is throwing error. Is it a bug then?

marklogic-client-api/src/main/java/com/marklogic/client/impl/SchemaDefImpl.java
anu3990: Will revisit this after the design is finalized.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
anu3990: Done.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
anu3990: Done

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
anu3990: i tried to send an invalid xml but the transaction does not go through

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
rjrudin: Hmm - after reading the package-info.java file and looking at the existing "Expr" classes, my thought is to stick with "Def" as the suffix. That's because the existing "Expr" classes seem different in nature - they all seem to be collections of functions, whereas our "Def" classes are defining sets of config options.

But as long as the 3 names are consistent, it's easy to attach "Expr" later if we decide that's the right move. Good question for Erik on Tuesday.

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
rjrudin: The client should still provide a logLevel if the user doesn't. However, see our chat room today - I don't think we're going to support this because it's unclear what "summary" and "detail" should even mean.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/RowManagerValidateDocTest.java
rjrudin: Check with @josvanroosmalen , he's rewriting this processor right now.

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
anu3990: okay. then i will revisit this as well :-)

PR #1423, opened 2022-10-31
Enabling test for bug 57987

rjrudin:
Want to see if this fails on Jenkins too.

Also improved mjsTransform test to include a write() call.


rjrudin:
Closing, just wanted to verify that this fails on Jenkins too

PR #1424, opened 2022-11-01
JAVA-242 : Tests for "validateDoc" pipeline operation

anu3990:
* OS and version

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
@@ -15,28 +15,23 @@
  */
 package com.marklogic.client.test.rows;
 
-import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.marklogic.client.datamovement.DataMovementManager;
 import com.marklogic.client.datamovement.WriteBatcher;
 import com.marklogic.client.document.DocumentManager;
+import com.marklogic.client.document.DocumentWriteSet;
 import com.marklogic.client.document.XMLDocumentManager;
 import com.marklogic.client.expression.PlanBuilder;
 import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.Format;
-import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.io.StringHandle;
-import com.marklogic.client.io.marker.AbstractWriteHandle;
 import com.marklogic.client.row.RowRecord;
 import com.marklogic.client.test.Common;
-import com.marklogic.client.type.PlanParamExpr;
 import org.junit.Before;
 import org.junit.Test;
-import org.springframework.util.StringUtils;
 
 import java.util.*;
 import java.util.stream.Collectors;
 
-import static com.marklogic.client.io.Format.JSON;
 import static com.marklogic.client.io.Format.XML;
 import static org.junit.Assert.*;
 
@@ -49,9 +44,12 @@ public class ValidateDocTest extends AbstractOpticUpdateTest {
     public void moreSetup(){
         dataMovementManager = Common.client.newDataMovementManager();
     }
-// TODO : test with rest-writer user when https://bugtrack.marklogic.com/57978 is fixed
+
     @Test
-    public void xmlSchema() {
+    public void testXmlSchemaWithLaxMode() {
+        if (!Common.markLogicIsVersion11OrHigher()) {
+            return;
+        }
         WriteBatcher writeBatcher = dataMovementManager.newWriteBatcher();
         DocumentMetadataHandle meta = newDefaultMetadata();
         dataMovementManager.startJob(writeBatcher);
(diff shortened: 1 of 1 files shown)


llinggit (approved):

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -51,7 +54,10 @@ public void moreSetup(){
     }
 // TODO : test with rest-writer user when https://bugtrack.marklogic.com/57978 is fixed
```
rjrudin, line 55: We don't want to change the user, we'll just remove the "xdmp-invoke" privilege from the "test-rest-writer" role.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -86,10 +92,69 @@ public void xmlSchema() {
         assertEquals(uriCountToWrite, persistedUris.size());
         //expectedUris.forEach(uri -> assertTrue("Did not find URI: " + uri, persistedUris.contains(uri)));
     }
+    @Test
+    public void testXmlSchemaWithStrictMode() {
```
rjrudin, line 96: It looks like only `testWithJsonSchemaAndInvalidJsons` is verifying that validation is occurring. The other tests are only verifying that "resultRows" is returning rows with expected URIs in them. I don't think this test class needs to care about that. It instead should follow a pattern of:

1. Use fromDocDescriptors (like `testWithJsonSchemaAndInvalidJsons` does) and pass in 2 docs, 1 that's valid and 1 that's invalid (only need 2)
2. Ensure that `write()` is part of the plan
3. Verify that `resultRows()` only returns 1 row
4. Verify that the valid doc is written and the invalid doc is not written

That's all that's needed for each of these tests. We don't need to test different kinds of invalid input, we just need to know that invalid input will be discarded, regardless of why it's invalid.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -86,10 +92,69 @@ public void xmlSchema() {
...
+                uris.add(rowRecord.getString("uri"));
+        }
+        // TODO : uncomment below after https://bugtrack.marklogic.com/57987 is fixed
+        // assertTrue(uris.size() == 4);
+        // assertTrue(uris.contains("/acme/1.xml"));
+        //assertTrue(uris.contains("/acme/2.xml"));
+        //assertTrue(uris.contains("/acme/3.xml"));
+        //assertTrue(uris.contains("/acme/0.xml"));
+
+        QueryManager queryManager = client.newQueryManager();
```
rjrudin, line 147: No need for these lines of code, the parent class will delete data at the start of each test.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -139,46 +204,108 @@ public void testUsingFromParamAndSchematron() {
         Iterator<RowRecord> rows = rowManager.resultRows(plan).iterator();
         while (rows.hasNext()){
             RowRecord rowRecord = rows.next();
+            System.out.println(rowRecord);
             if(rowRecord.getString("rowId")!=null)
                 expectedUris.add(rowRecord.getString("rowId"));
         }
-        // TODO: uncomment the below after https://bugtrack.marklogic.com/58055 is fixed
-        // assertTrue(expectedUris.size() == 1);
+        // TODO : uncomment below after https://bugtrack.marklogic.com/57978 is fixed
```
rjrudin, line 211: These tests can just verify that the expected number of rows are returned. And by following the pattern above, where only 1 row is expected to be returned, none of these tests will be impacted by bug 57978 (which only affects result sets with 2 or more rows).

rjrudin (changes requested):

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
anu3990: Hi @rjrudin , the "testUsingFromParamAndSchematron" test is also verifying something similar. Out of the the 3 documents sent as attachment only one (doc3.xml) should be returned. (I think currently it is failing due to the xdmp-update permission issue). Do you recommend to add a write operation to it?

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
anu3990: Actually, these docs were not removed after running all the tests. Was it because a different user was used? (line 105). Also, i had to use the rest-writer because the user test-rest-writer was giving exception while inserting docs, something like  - "Atleast one update permission needed while inserting documents"

rjrudin: That's a good exception! We don't want to rely on users having default permissions because it can hide actual problems. Just use "newDefaultMetadata()" (which is in AbstractOpticUpdateTest) to construct a `DocumentMetadataHandle`, it has read/update permissions on it. 

Also, the docs should not be removed after a test completes - that makes debugging failed tests much easier. Instead, docs should be removed at the start of a test, which the parent class does by deleting everything in the "/acme/" directory.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
rjrudin: Yes, I think `write` is good to do for all of these tests to make them more realistic, as I think well over 90% of the time that `validateDoc` is used, it'll be right before docs are written.

anu3990: Hi Rob, left this test unchanged and added a separate test "testWriteWithXmlSchemaWithStrictMode" for testing write.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -136,49 +221,92 @@ public void testUsingFromParamAndSchematron() {
                 "</user>").withFormat(Format.XML));
         plan = plan.bindParam(param, new JacksonHandle(array), Collections.singletonMap("doc", attachments));
 
+
+        List<RowRecord> rows = resultRows(plan);
+        assertEquals(1, rows.size());
```
anu3990, line 226: @rjrudin is this test missing some binding? I get this error - 
Local message: failed to apply resource at rows: Bad Request. Server Message: OPTIC-INVALARGS: fn.error(null, 'OPTIC-INVALARGS', 'missing or empty URI value in write()'); -- Invalid arguments: missing or empty URI value in write()

rjrudin: Yes - I recommend using `bindParam(String, DocumentWriteSet)` as shown in `xmlDocumentWriteSet` in `FromParamWriteTest`. I think that's the easiest way to setup a row with URI + doc + permissions. In fact, for writing documents, I think a user will always want to use a `DocumentWriteSet` or sequence of `DocumentWriteOperation`s because it's otherwise a bit of a pain to get the URI + doc + permissions columns together.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
rjrudin: The "lax" test should still have at least one doc that fails. Right now, I can comment out the line for `validateDoc` and the test still passes, meaning that that test isn't really doing anything yet - it's only testing that `fromDocUris` and `joinDoc` work, but we already have separate tests for those. Same is true for the "strict" test here - since all of the docs are valid, we don't know if the validation is actually working or not.

PR #1432, opened 2022-11-03
Added more transformDoc and remove tests

rjrudin:
Also added rest-tracer for easier debugging with the test user.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/TransformDefImpl.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/TransformDefImpl.java
@@ -1,12 +1,14 @@
 package com.marklogic.client.impl;
 
-import java.util.HashMap;
-import java.util.Map;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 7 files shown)


rjrudin:
Addresses at least bugs 58101 and 58102

marklogic-client-api/src/main/java/com/marklogic/client/impl/TransformDefImpl.java
```diff
@@ -24,11 +26,36 @@ public StringBuilder exportAst(StringBuilder strb) {
         node.put("path", path);
         node.put("kind", kind);
         if (params != null) {
-            node.putPOJO("params", params);
+            node.putPOJO("params", prepareParamsDuringExport());
         }
         return strb.append(node);
     }
 
+    private Map<String, Object> prepareParamsDuringExport() {
```
rjrudin, line 34: @anu3990  @llinggit  @josvanroosmalen  I couldn't find a better way to do this. The user's map can contain a variety of values - strings, numbers, booleans, ObjectNode's, and now PlanColumn's. So Jackson's `putPOJO` method is very convenient. But that ran into a stack overflow error on the `PlanColumn` class. So I worked around this by first serializing the `PlanColumn` into a JSON string and then converting that back into an `ObjectNode`. Everything works fine with this approach, just seems hacky.

PR #1433, opened 2022-11-04
Several fixes for functional tests

rjrudin:
The two "stop job" tests are still failing with off-by-one errors; I just added comments for those. The others were all easy fixes. The geospatial value changes are all due to an expected change in the server last night.


georgeajit (commented):
toWKT(); will this work with 10.0 ML build and develop?

anu3990 (approved):

rjrudin:
@georgeajit  It won't work, I'm going to see which tests fail on ML 10 next and then disable them when the ML version is 10. We're taking the same approach for the marklogic-client-api tests.

PR #1437, opened 2022-11-07
Removed unused argument in Generator

rjrudin:
This was producing a warning during the build process, which makes the Jenkins logs a bit harder to debug. Trivial to remove.


BillFarber (approved):
Seems fine to me, but I don't know Kotlin.

PR #1439, opened 2022-11-07
Can now run some functional tests on test app

rjrudin:
Added files to the test app so that functional tests can run against port 8014 with their own database. 

Added `AbstractFunctionalTest` as a base class for tests that will use the test app instead of spending a minute or more setting up and tearing down their own test app. That class also resolves a lot of duplication across at least the Optic functional tests. 

Removed some unused stuff as well.


georgeajit (approved):
Thank you for the refactoring.

PR #1443, opened 2022-11-07
Updating the Jenkins file to add new enhancements and fixing some bugs

SameeraPriyathamTadikonda:


Jenkinsfile
```diff
@@ -1,5 +1,7 @@
 @Library('shared-libraries') _
 
+def Email = "rob.rudin@marklogic.com"
```
rjrudin, line 3: Is there a way to do this via an environment variable? I'd rather not have my work email address sitting around like this.

Jenkinsfile
```diff
@@ -64,7 +76,7 @@ pipeline{
           export GRADLE_USER_HOME=$WORKSPACE/$GRADLE_DIR
           export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
           cd java-client-api
-          ./gradlew mlDeploy
+          ./gradlew mlDeploy -PmlForestDataDirectory=/space
```
rjrudin, line 79: I added this in a separate PR where was I fixing a config issue with the functional tests

Jenkinsfile
SameeraPriyathamTadikonda: Sure, I'll pass it as a parameter.

rjrudin (approved):

PR #1452, opened 2022-11-10
Moved all converted tests to "fastfunctest"

rjrudin:
The changes in each file are doing to running "Optimize imports" in Intellij on each class, as most had several unused imports or more. 

One functional change - tweaked `TestEvalXquery` so that not being in the PCT timezone doesn't cause the test to fail. 

Another - modified testValueConstraintWildCard to not assume an order of results since there's no sort operator. 

And - fixed testBug22179, which had constraint-name and text out of order in the search options.


BillFarber (approved):
blind approval

PR #1462, opened 2022-11-16
Updated and standardized copyright for 2022

rjrudin:
@anadisharmaa  Can you run your copyright detection tool on this branch and let me know the results? Note the 14 .txt files I mentioned to you - those are src/test files that aren't part of a released artifact and cannot easily have a copyright added to them because it'll break the tests that depend on them.


georgeajit (approved):
Went through all diffs. They look fine to me.

Issue #1336, opened 2022-02-22
Generate code for 3rd parameter of op.xpath


llinggit:
Need to modify optic-defs.json under xdmp/src/makefiles/optic first.
Then run code generation again to refresh the change in Java Client API

So we can address your issue, please include the following:
### Version of MarkLogic Java Client API
See Readme.txt

### Version of MarkLogic Server

See admin gui on port 8001 or run xdmp:version() in Query Console - port 8000)

### Java version

Run `java -version`

### OS and version

For MAC, run `sw_vers`. \
For Windows, run  `systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`\
For Linux, run `cat /etc/os-release` and `uname -r`

rjrudin:
We have a PR for this; we're tracking items in JIRA, so closing this as we don't need it to be tracked in github.

PR #1463, opened 2022-11-16
Generate code for 3rd parameter of op.xpath

anu3990:
* What issue are you addressing with this pull request? https://github.com/marklogic/java-client-api/issues/1336 
* OS and version


marklogic-client-api/src/main/java/com/marklogic/client/impl/XdmpExprImpl.java
```diff
@@ -1080,6 +1080,45 @@ public ServerExpression type(ServerExpression value) {
   }
 
   
+  @Override
+  public ServerExpression unquote(ServerExpression arg) {
+    if (arg == null) {
+      throw new IllegalArgumentException("arg parameter for unquote() cannot be null");
+    }
+    return new BaseTypeImpl.DocumentNodeSeqCallImpl("xdmp", "unquote", new Object[]{ arg });
+  }
```
anu3990, line 1089: @rjrudin and @llinggit , the code currently does not compile since bringing back "unquote" requires us to add a new "DocumentNodeSeqCallImpl" in BaseTypeImpl. Any suggestions here?

rjrudin: That seems fine to me, since BaseTypeImpl isn't generated. Though, do we just assume that DocumentNodeSeqCallImpl should have a constructor of `String, String, Object[]`?

rjrudin (approved):
The 2 test failures are intermittent Bulk DS failures, so approving

PR #1470, opened 2022-11-29
Prototype of Java 17 support

rjrudin:
Most of this is just replacing javax.xml with jakarta.xml. We do need to use a different version of jakarta.xml.bind though based on whether Java 8 or 11+ is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,6 @@ build
 buildSrc/build
 
 .classpath
-.gitignore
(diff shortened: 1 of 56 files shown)


rjrudin:
Closing this for now, as I think we just need a README update for now. Bumping the Java Client to depend on jakarta.xml.bind would still be a good thing, but it's not urgent.

PR #1471, opened 2022-11-30
Updated README to address how to use Java Client on Java 11 and 17

rjrudin:
This addresses a gap in the docs where we've never explained how to include JAXB if a user is using Java 11 or 17 and wishes to use the JAXB support in the Java Client. 

Can also now build and run all the tests using Java 11 or 17 (we've only used Java 8 before). That simply required the following:

- Don't fail the build on javadoc errors; Java 11 is much more particular about these, and we ought to do a cleanup of these errors soon, but we can ignore them for now
- Bumped up Gradle wrapper version to the latest 7.x; to build and run on Java 17, need at least Gradle 7.3
- Bumped up Mockito from the very, very old mockito-all dependency (last updated in 2015) to the latest ones; this fixed a simple problem with Mockito that popped up when testing with Java 11 and Java 17
- Added a duplicatesStrategy for ml-development-tools to address an issue where the plugin properties file was getting copied over twice; this occurred as a result of upgrading from Gradle 6 to 7. There is likely a better way to handle this, but this approach (which the error message recommends using) works for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ implemented by an endpoint on the server.
 
 ## QuickStart
 
-To use the API in your maven project, include the following in your pom.xml:
+To use the API in your [Maven](https://maven.apache.org/) project, include the following in your pom.xml file:
 
     <dependency>
         <groupId>com.marklogic</groupId>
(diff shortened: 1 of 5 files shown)


BillFarber (approved):

rjrudin:
This shows an example of a test that fails on Java 11 and 17 if the old JAXB dependencies are not included, but it runs correctly with those dependencies present - https://github.com/marklogic-community/ml-javaclient-util/pull/164

anu3990 (approved):

PR #1481, opened 2022-12-08
DEVO-690: Pass Java version as Parameter

SameeraPriyathamTadikonda:


rjrudin (changes requested):
@SameeraPriyathamTadikonda  Error in Jenkins:

Z] WorkflowScript: 4: expecting ')', found '=' @ line 4, column 24.

[2022-12-08T21:55:15.546Z]        if(env.JAVA_VERSION="JAVA17"){

Guessing it has to be "==".

PR #1494, opened 2022-12-20
DEVEXP-148: Extracted AuthenticationConfigurer to simplify auth code

rjrudin:
Only doing this for basic/digest as we have sufficient tests for both of those authentication types. Intent is to both make this part of OkHttpServices more readable and to make it easier to add token-based auth (i.e. via a separate class instead of stuffing it into OkHttpServices). 

I did refactor some of the kerberos/cert/saml code, as all 5 auth branches were needlessly doing null checks on getSSLContext and getTrustManager. `configureSocketFactory` is already checking to see if those are null or not before doing anything with them.

Added `getTrustManager()` to `SecurityContext` to simplify accessing this too - the code for accessing this was needlessly complicated given that all implementations of `SecurityContext` already had this public method.


marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -325,6 +325,13 @@ public interface SecurityContext {
      */
     SSLContext getSSLContext();
 
+	  /**
+	   * @return the trust manager for this connection
+	   * @since 6.1.0 ; was already implemented by both implementations of this interface, and exposing it here
```
rjrudin, line 330: For each new public addition going forward, we're going to use `@since` to make life easier for all of us.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -203,71 +199,52 @@ private FailedRequest extractErrorFields(Response response) {
...
 
-    OkHttpClient.Builder clientBldr = newClientBuilder();
+	  if (authenticationConfigurer != null) {
+		  authenticationConfigurer.configureAuthentication(clientBuilder, securityContext);
+	  }
 
-	if (securityContext instanceof BasicAuthContext) {
-	    BasicAuthContext basicContext = (BasicAuthContext) securityContext;
-	    if (basicContext.getSSLContext() != null) {
-            sslContext = basicContext.getSSLContext();
```
rjrudin, line 218: All of this null-checking on sslContext/trustManager was unnecessary, so I removed it. configureSocketFactory already checks to see if these are not null before doing anything with them.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -203,71 +199,52 @@ private FailedRequest extractErrorFields(Response response) {
...
-	    }
-	}
+	  SSLHostnameVerifier sslVerifier = null;
+	  if (sslContext != null || securityContext instanceof CertificateAuthContext) {
+		  sslVerifier = securityContext.getSSLHostnameVerifier() != null ?
+			  securityContext.getSSLHostnameVerifier() :
+			  SSLHostnameVerifier.COMMON;
+	  }
 
-    this.configureSocketFactory(clientBldr, sslContext, trustManager);
```
rjrudin, line 269: I am not a fan of removing vowels from words. :)

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -316,33 +293,31 @@ private OkHttpClient.Builder newClientBuilder() {
...
-                    throw new IllegalArgumentException("no trust manager and cannot initialize factory for default", e);
-                } catch (NoSuchAlgorithmException e) {
-                    throw new IllegalArgumentException("no trust manager and no algorithm for default manager", e);
-                } catch (KeyManagementException e) {
-                    throw new IllegalArgumentException("no trust manager and cannot initialize context with default", e);
-                }
-            } else {
-                clientBuilder.sslSocketFactory(new SSLSocketFactoryDelegator(sslContext.getSocketFactory()), trustManager);
-            }
+        } else if (trustManager != null) {
```
rjrudin, line 296: I reworked this a bit to do the "simple" conditions first - life is simple if sslContext is null or if both sslContext and trustManager are not null. Otherwise, things get complicated - we have to build up a default trustManager.

anu3990 (approved):

PR #1495, opened 2022-12-20
DEVEXP-144: DatabaseClient now supports basePath

rjrudin:
The app changes were fairly simple here, with the most notable part being the addition of HttpUrlBuilder to encapsulate the logic for building a base URL, which has to account for basePath now. Added a bunch of unit tests for this too. 

Most of the rest of this PR is test stuff, specifically the new support for setting up a reverse proxy server and configuring the tests to use that server instead of talking to ML directly.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
@@ -42,6 +42,7 @@ import com.marklogic.client.eval.EvalResultIterator;
 import com.marklogic.client.impl.okhttp.BasicAuthenticationConfigurer;
 import com.marklogic.client.impl.okhttp.DigestAuthenticationConfigurer;
 import com.marklogic.client.impl.okhttp.AuthenticationConfigurer;
+import com.marklogic.client.impl.okhttp.HttpUrlBuilder;
 import com.marklogic.client.io.*;
 import com.marklogic.client.io.marker.*;
 import com.marklogic.client.query.*;
@@ -198,7 +199,7 @@ public class OkHttpServices implements RESTServices {
   }
 
   @Override
-  public void connect(String host, int port, String database, SecurityContext securityContext){
+  public void connect(String host, int port, String basePath, String database, SecurityContext securityContext){
 
 	  if (host == null)
 		  throw new IllegalArgumentException("No host provided");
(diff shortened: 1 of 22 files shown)


test-app/src/main/java/com/marklogic/client/test/ReverseProxyServer.java
```diff
@@ -0,0 +1,142 @@
...
+            if (args.length > 1) {
+                markLogicHost = args[1];
+                if (args.length > 2) {
+                    serverPort = Integer.parseInt(args[2]);
+                }
+            }
+        }
+
+        Map<String, URI> mapping = new HashMap<>();
+        mapping.put("unitTest", new URI("http://" + markLogicHost + ":8012"));
```
rjrudin, line 54: Rob - do test/unit here

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -247,13 +248,7 @@ public void connect(String host, int port, String database, SecurityContext secu
...
 
     this.database = database;
-
-    this.baseUri = new HttpUrl.Builder()
-      .scheme(sslContext == null ? "http" : "https")
-      .host(host)
-      .port(port)
-      .encodedPath("/v1/ping")
-      .build();
+    this.baseUri = HttpUrlBuilder.newBaseUrl(host, port, basePath, sslContext);
```
rjrudin, line 251: This is the main change here, along with what's below - all the stuff in HttpUrlBuilder is what really matters.

test-app/src/main/ml-config/servers/java-functest.json
```diff
@@ -1,12 +1,13 @@
...
+	"port": 8014,
+	"server-type": "http",
+	"root": "/",
+	"content-database": "java-functest",
+	"modules-database": "%%MODULES_DATABASE%%",
+	"default-error-format": "json",
+	"error-handler": "/MarkLogic/rest-api/error-handler.xqy",
+	"url-rewriter": "/MarkLogic/rest-api/rewriter.xml",
+	"rewrite-resolves-globally": true,
+	"authentication": "digestbasic"
```
rjrudin, line 12: digestbasic is used now so that both digest and basic auth will work, as the reverse proxy only supports basic.

anu3990 (approved):

BillFarber (approved):

PR #1498, opened 2022-12-28
Migrating to JUnit 5

rjrudin:
Nearly every test was touched here due to the change in import. 

Another source of changes was accounting for JUnit 5's move of "String message" from being the first arg to the lazy arg. All existing  assertion messages (which was nearly all of them) simply restated what the assertion itself already made clear and thus were redundant, and so they were removed.

Upgraded xmlunit to 2.9.0 but still using the "legacy" APIs to minimize changes to any tests using xmlunit. 

No application code is changed by this.


BillFarber (approved):
Just did spot-checking of the tests.
LGTM

PR #1501, opened 2022-12-29
DEVEXP-143: Now supporting authentication for MarkLogic Cloud

rjrudin:
Only downside is there's not yet a way to do this in an automated test against a real cloud instance. So relying on small unit tests and manual testing for now.


BillFarber (approved):
LGTM.
However, I'm not familiar with cloud authentication using tokens.

PR #1504, opened 2022-12-29
Bumping Jackson to 2.14.1

rjrudin:
Also changed all dependencies to use the shorter syntax instead of including the words group/name/version. 

Bumped logback to 1.3.5 as well. And mockito to 4.10.0. 

And bumped json-schema-validator to 1.0.76 so that it also is using Jackson 2.14.


BillFarber (approved):
LGTM once the last test passes.

PR #1507, opened 2023-01-03
Added phase to Jenkinsfile for running all tests with reverse proxy

rjrudin:
Removed email notifications for now - we're not making use of them.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,11 +10,12 @@ def getJava(){
     }
 }
 
-def runtests(String type, String version){
+def runAllTests(String type, String version, Boolean useReverseProxy){
             copyRPM type, version
             setUpML '$WORKSPACE/xdmp/src/Mark*.rpm'
             copyConvertersRPM type,version
             setUpMLConverters '$WORKSPACE/xdmp/src/Mark*Converters*.rpm'
+
             sh label:'deploy test app', script: '''#!/bin/bash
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -10,26 +10,40 @@ def getJava(){
...
-            sh label:'run marklogic-client-api tests', script: '''#!/bin/bash
-                export JAVA_HOME=$JAVA_HOME_DIR
-                export GRADLE_USER_HOME=$WORKSPACE/$GRADLE_DIR
-                export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
-                cd java-client-api
-                mkdir -p marklogic-client-api/build/test-results/test
-                ./gradlew marklogic-client-api:test  || true
-            '''
+
+            if (useReverseProxy) {
```
rjrudin, line 27: @SameeraPriyathamTadikonda  Let me know if this looks good before I merge this. Also, it would be nice to avoid the duplication of the 3 export lines. I assume that could be done with a new function, where we pass it the text we want to execute, and the function would prepend the export statements?

SameeraPriyathamTadikonda: @rjrudin  I'll see if there are any other plugins to inject environment variables to the shell. right now, we cannot use a groovy function to export variables in a shell.

SameeraPriyathamTadikonda (approved):

PR #1517, opened 2023-01-13
Cleanup of unit test plumbing

rjrudin:
Did the following:

1. Renamed "adminClient" to "restAdminClient" so I stop thinking that "admin client" refers to an "admin client" (that represents most of the changes here)
2. Reduced the number of overloaded "makeNewClient" methods in Common.java

No functional changes here.


BillFarber (approved):
LGTM

PR #1521, opened 2023-01-23
Forcing port to 443 when using MarkLogic cloud authentication

rjrudin:
443 is written in stone for now, and thus the UX can be simpler by forcing the port to that value.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/extra/okhttpclient/OkHttpClientBuilderFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/extra/okhttpclient/OkHttpClientBuilderFactory.java
@@ -28,7 +28,7 @@ import okhttp3.OkHttpClient;
  */
 public interface OkHttpClientBuilderFactory {
 
-	static OkHttpClient.Builder newOkHttpClientBuilder(String host, int port, DatabaseClientFactory.SecurityContext securityContext) {
-		return OkHttpUtil.newOkHttpClientBuilder(host, port, securityContext);
+	static OkHttpClient.Builder newOkHttpClientBuilder(String host, DatabaseClientFactory.SecurityContext securityContext) {
+		return OkHttpUtil.newOkHttpClientBuilder(host, securityContext);
 	}
 }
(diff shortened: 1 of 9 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/extra/okhttpclient/OkHttpClientBuilderFactory.java
```diff
@@ -28,7 +28,7 @@
  */
 public interface OkHttpClientBuilderFactory {
 
-	static OkHttpClient.Builder newOkHttpClientBuilder(String host, int port, DatabaseClientFactory.SecurityContext securityContext) {
-		return OkHttpUtil.newOkHttpClientBuilder(host, port, securityContext);
+	static OkHttpClient.Builder newOkHttpClientBuilder(String host, DatabaseClientFactory.SecurityContext securityContext) {
```
rjrudin, line 31: This is what ml-app-deployer users - it becomes simpler because the ML cloud auth configurer can safely assume usage of 443 for obtaining an access token, so "port" doesn't need to be provided anymore.

marklogic-client-api/src/test/java/com/marklogic/client/impl/okhttp/MarkLogicCloudAuthenticationConfigurerTest.java
```diff
@@ -17,31 +17,21 @@ public class MarkLogicCloudAuthenticationConfigurerTest {
...
-		HttpUrl tokenUrl = new MarkLogicCloudAuthenticationConfigurer("somehost", 443).buildTokenUrl(
+		HttpUrl tokenUrl = new MarkLogicCloudAuthenticationConfigurer("somehost").buildTokenUrl(
 			new DatabaseClientFactory.MarkLogicCloudAuthContext("doesnt-matter")
 				.withSSLContext(SSLContext.getDefault(), null)
 		);
 		assertEquals("https://somehost/token", tokenUrl.toString(),
 			"When the port is 443, OkHttp won't include it in the URL");
 	}
 
-	@Test
```
rjrudin, line 28: This test reflects a now-invalid scenario and has thus been removed.

rjrudin:
@anu3990 This is likely worth doing in the Node Client as well.

BillFarber (approved):

PR #1522, opened 2023-01-24
Added more builder methods

rjrudin:
Syntatic sugar for when the client knows the desired security context type.

### DIFF
--- a/marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/AbstractFunctionalTest.java
+++ b/marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/AbstractFunctionalTest.java
@@ -57,7 +57,7 @@ public abstract class AbstractFunctionalTest extends BasicJavaClientREST {
         isML11OrHigher = version.getMajor() >= 11;
 
 		client = newDatabaseClientBuilder().build();
-		schemasClient = newClientForDatabase("java-functest-schemas");
+		schemasClient = newDatabaseClientBuilder().withDatabase("java-functest-schemas").build();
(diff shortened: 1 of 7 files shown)


marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/AbstractFunctionalTest.java
```diff
@@ -57,7 +57,7 @@ public static void initializeClients() throws Exception {
         isML11OrHigher = version.getMajor() >= 11;
 
 		client = newDatabaseClientBuilder().build();
-		schemasClient = newClientForDatabase("java-functest-schemas");
```
rjrudin, line 60: `newClientForDatabase` was only used in 2 places, which doesn't seem like justification enough for a helper method since it barely saves on any code.

BillFarber: Would it be useful to the users of the client?

BillFarber (approved):

marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/AbstractFunctionalTest.java
rjrudin: `newClientForDatabase` was in src/test/java. I'm trying to slim down `ConnectedRestQA` as much as possible so it's easier to comprehend. So if there's a helper method only used in one or two places that doesn't help much at all, I'm removing those. No impact on Java Client users.

PR #1525, opened 2023-01-26
Enhancements for connecting to ML Cloud

rjrudin:
These fixes are driven by testing against the ReverseProxyServer which can now emulate ML Cloud. The fixes:

1. If user chooses an sslProtocol of "default", then the JVM's default trust manager will be used as well.
2. If the user uses `withMarkLogicCloudAuth`, then the SSL protocol will default to "default".
3. The call to "/token" in ML Cloud now uses the user-provided SSL hostname verifier (this was not required on the ML Cloud instance I tested with, but that is not guaranteed to be true).


BillFarber (approved):
LGTM

anu3990 (approved):

PR #1538, opened 2023-03-27
Updated Gradle wrapper jar and scripts

rjrudin:
Was using an older jar that was flagged with a security vulnerability.


BillFarber (approved):

rjrudin:
There's a test failing due to an unrelated change in the server, will submit a separate PR for that.

PR #1554, opened 2023-04-11
Trying MarkLogic 11.1 in Jenkinsfile

rjrudin:


Jenkinsfile
```diff
@@ -154,9 +154,9 @@ pipeline{
         }
       }
       steps {
-        copyRPM 'Latest','11.0'
+        copyRPM 'Latest','11.1'
```
SameeraPriyathamTadikonda, line 157: @rjrudin  library will still pull 11.1 images(latest images) even if we pass 11.0. I need to update the library first before updating the repositories. I'll update all at once. We can close the PR until then

PR #1559, opened 2023-05-09
Bumping OkHttp to 4.11, Jackson to 2.14.3

rjrudin:
And a few other patch upgrades. Also set version to 6.2-SNAPSHOT in preparation for a 6.2.1 release.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group=com.marklogic
-version=6.3-SNAPSHOT
+version=6.2-SNAPSHOT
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=com.marklogic
-version=6.3-SNAPSHOT
+version=6.2-SNAPSHOT
```
BillFarber, line 2: The version number went backward?

BillFarber (approved):
Assuming the version number is correct.

PR #1560, opened 2023-05-10
Trying out Jackson 2.15.0

rjrudin:


rjrudin:
Holding off on this for now until this can be tested with all the connectors that depend on it.

PR #1562, opened 2023-05-10
DEVEXP-417 Tweaked token renewal to be based on 401 instead of 403

rjrudin:


BillFarber (commented):
Looks good to me, but it failed checks, so not approving for now.

rjrudin:
That's fine, the failures were just because of the internal devops problem yesterday with SSL.

BillFarber (approved):

PR #1566, opened 2023-05-17
DEVEXP-460 resultRows now honors point-in-time query timestamp

rjrudin:


BillFarber (dismissed):

rjrudin:
Had to make the test delete the one document it inserts since not all the tests are "proper" in terms of clearing data before the start of each test. Sigh.

BillFarber (approved):

PR #1567, opened 2023-05-25
DEVEXP-471 : ValidateDoc does not receive accurate response when schema or document does not exist

anu3990:
* OS and version


rjrudin (approved):

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/ValidateDocTest.java
```diff
@@ -294,4 +294,15 @@ public void testWithJsonSchemaAndInvalidJsons() {
...
+	@Test
+	public void testWithNonExistingSchema() {
+		PlanBuilder.ModifyPlan plan = op
+			.fromDocDescriptors(
+				op.docDescriptor(newWriteOp("/acme/doc1.json", mapper.createObjectNode().put("count", 1).put("total",2)))
+			).validateDoc(op.col("doc"),
+				op.schemaDefinition("jsonSchema").withSchemaUri("/validateDoc/Non-Existing.json"));
+		rowManager.execute(plan.write());
+		DocumentManager docMgr = Common.client.newDocumentManager();
+		assertTrue(docMgr.exists("/acme/doc1.json") == null);
```
rjrudin, line 306: We should include an explanation here since the behavior is a little surprising. Also, `assertNull` is more appropriate to use. How about:

assertNull(docMgr.exists("..."), "When a schema is used that doesn't exist, the validation by the server is expected to silently fail, and thus the document is not written"); 

Writing that out does make the behavior seem pretty surprising - it really sounds like an error should be thrown instead, but that would be a change to the server.

rjrudin (changes requested):

rjrudin (approved):

PR #1570, opened 2023-06-13
Added test for a custom X509TrustManager

rjrudin:
Figured out the plumbing for doing this with the Manage API, figured I should document it via a test here.

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/junit5/RequireSSLExtension.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/junit5/RequireSSLExtension.java
@@ -5,9 +5,16 @@ import com.marklogic.client.test.Common;
 import com.marklogic.mgmt.ManageClient;
 import com.marklogic.mgmt.resource.appservers.ServerManager;
 import com.marklogic.mgmt.resource.security.CertificateTemplateManager;
+import com.marklogic.rest.util.Fragment;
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/junit5/RequireSSLExtension.java
```diff
@@ -52,6 +59,39 @@ public void afterAll(ExtensionContext context) {
...
+				return new X509Certificate[]{getCertificate()};
+			}
+		};
+	}
+
+	private static X509Certificate getCertificate() {
+		CertificateTemplateManager mgr = new CertificateTemplateManager(Common.newManageClient());
+
+		Fragment response = mgr.getCertificatesForTemplate(TEMPLATE_NAME);
+		String cert = response.getElementValue("/msec:certificate-list/msec:certificate/msec:pem");
```
rjrudin, line 87: @BillFarber Could reproduce this in mlxprs if needed, though I don't know how to make a certificate / trust manager in Node or what the equivalent is. But this is the key part - after generating a temporary certificate, you can download it and get the pem via the above code.

BillFarber: I've been thinking that we need to add a test or two to exercise certs, but I haven't written any stories for it.

BillFarber (approved):

PR #1571, opened 2023-07-18
DEVEXP-508 Certificate auth no longer requires a file path

rjrudin:
`DatabaseClientBuilder` now supports the constructor that accepts an `SSLContext` and `X509TrustManager` without requiring a certificate file and password.


marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientBuilder.java
```diff
@@ -158,6 +158,12 @@ public DatabaseClientBuilder withCertificateAuth(String file, String password) {
 			.withCertificatePassword(password);
 	}
 
+	public DatabaseClientBuilder withCertificateAuth(SSLContext sslContext, X509TrustManager trustManager) {
```
rjrudin, line 161: This was not needed for the fix, it's just some syntactic sugar for the use case of doing certificate auth with a user-provided SSLContext instead of a certificate file path.

marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
```diff
@@ -210,15 +210,19 @@ private DatabaseClientFactory.SecurityContext newCloudAuthContext() {
...
+			try {
+				if (password != null && password.trim().length() > 0) {
+					return new DatabaseClientFactory.CertificateAuthContext(file, password, sslInputs.getTrustManager());
+				}
+				return new DatabaseClientFactory.CertificateAuthContext(file, sslInputs.getTrustManager());
+			} catch (Exception e) {
+				throw new RuntimeException("Unable to create CertificateAuthContext; cause " + e.getMessage(), e);
+			}
 		}
+		return new DatabaseClientFactory.CertificateAuthContext(sslInputs.getSslContext(), sslInputs.getTrustManager());
```
rjrudin, line 225: This is the actual fix - if the user doesn't provide a certificate file, then use the SSLContext and X509TrustManager that they provided.

BillFarber (approved):

PR #1593, opened 2023-08-25
DEVEXP-562 Can now prevent Accept-Encoding=gzip header being used

rjrudin:
DatabaseClientFactory also supports multiple configurators now. This was essential for testing this new story, but it also aligns with what a user would expect a method named "add" to do (as opposed to "set"). `clearConfigurators` was also added both for testing purposes and so a user could remove all the configurators in case one or more should no longer be used.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
@@ -63,7 +63,7 @@ import com.marklogic.client.io.marker.ContentHandleFactory;
  */
 public class DatabaseClientFactory {
 
-  static private ClientConfigurator<?> clientConfigurator;
+  static private List<ClientConfigurator<?>> clientConfigurators = new ArrayList<>();
   static private HandleFactoryRegistry handleRegistry =
     HandleFactoryRegistryImpl.newDefault();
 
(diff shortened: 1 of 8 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -1408,17 +1413,19 @@ static public DatabaseClient newClient(String host, int port, String basePath, S
...
-          }
+      if (clientConfigurators != null) {
+		  clientConfigurators.forEach(configurator -> {
+			  if (configurator instanceof OkHttpClientConfigurator) {
+				  OkHttpClient okHttpClient = (OkHttpClient) services.getClientImplementation();
+				  OkHttpClient.Builder clientBuilder = okHttpClient.newBuilder();
+				  ((OkHttpClientConfigurator) configurator).configure(clientBuilder);
+				  ((OkHttpServices) services).setClientImplementation(clientBuilder.build());
+			  } else if (configurator instanceof HttpClientConfigurator) {
+				  // do nothing as we no longer use HttpClient so there's nothing this can configure
```
BillFarber, line 1424: Does this mean somebody might configure an HttpClientConfigurator object and expect it to do something, but then nothing happens?

rjrudin: Correct, that's been the behavior since release 4.2.0. In fact, I'll submit a separate PR to mark that interface as deprecated so it can be removed in 7.0, along with this block of code.

BillFarber (approved):

PR #1607, opened 2023-10-02
Refactored logic for determining SSL inputs

rjrudin:
I was looking at this while analyzing support for 2-way SSL and I couldn't really understand what I wrote. So I refactored in the following way:

1. All construction for SSLContext/TrustManager is in buildSSLInputs.
2. There are 4 approaches, each clearly identified and implemented in its own method.
3. I moved the tests I use for verifying SSL support to a "test.ssl" package (no changes to the tests themselves).

This may actually fix a bug but I'm not certain. The bug would have been that "newCertificateAuthContext" was called before all the SSL-input logic had occurred, meaning that e.g. "default" as an sslProtocol value would not have impacted certificate authentication.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
@@ -19,8 +19,6 @@ import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.DatabaseClientBuilder;
 import com.marklogic.client.DatabaseClientFactory;
 import com.marklogic.client.extra.okhttpclient.RemoveAcceptEncodingConfigurator;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.X509TrustManager;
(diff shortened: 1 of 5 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
```diff
@@ -329,61 +271,124 @@ private DatabaseClientFactory.SSLHostnameVerifier determineHostnameVerifier() {
 	 * X509TrustManager.
 	 *
 	 * @param authType used for applying "default" as the SSL protocol for MarkLogic cloud authentication in
-	 *                            case the user does not define their own SSLContext or SSL protocol
+	 *                 case the user does not define their own SSLContext or SSL protocol
 	 * @return
 	 */
 	private SSLInputs buildSSLInputs(String authType) {
-		SSLContext sslContext = null;
+		X509TrustManager userTrustManager = getTrustManager();
```
rjrudin, line 278: This is where all the changes start. This method should be a lot easier to understand now.

marklogic-client-api/src/main/java/com/marklogic/client/impl/SSLUtil.java
```diff
@@ -25,7 +28,12 @@
 public interface SSLUtil {
 
 	static X509TrustManager getDefaultTrustManager() {
-		return (X509TrustManager) getDefaultTrustManagers()[0];
+		X509TrustManager trustManager = (X509TrustManager) getDefaultTrustManagers()[0];
+		Logger logger = LoggerFactory.getLogger(SSLUtil.class);
+		if (logger.isDebugEnabled() && trustManager.getAcceptedIssuers() != null) {
```
rjrudin, line 33: I moved this logging here as it's useful to see for debugging purposes regardless of when this method is called.

BillFarber (approved):
LGTM

PR #1612, opened 2023-10-24
Removing RequiresML11

rjrudin:
Haven't really needed this since Java Client 6.0 was released, as that was associated with MarkLogic 11. We don't have any Jenkins jobs running against MarkLogic 10 either (we would do that if we do a new 5.x release). 

This simplifies using the Optic code generator by not modifying PlanGeneratedTest anymore.


rjrudin:
Ack, holding off on this for now as Jenkinsfile runs the tests against 10.0-9.

PR #1614, opened 2023-10-24
DEVEXP-585 Can now configure 2-way SSL

rjrudin:
Bumped the version to 6.4-SNAPSHOT as well so we can start using it in the Mule connector. 

The main value here is in TwoWaySSLTest, which can use these new properties instead of having to construct an SSLContext itself.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
@@ -151,7 +151,7 @@ public class DatabaseClientPropertySource {
 		}
 		final String authType = (String) typeValue;
 
-		final SSLInputs sslInputs = buildSSLInputs(authType);
+		final SSLUtil.SSLInputs sslInputs = buildSSLInputs(authType);
 		DatabaseClientFactory.SecurityContext securityContext = newSecurityContext(authType, sslInputs);
 		if (sslInputs.getSslContext() != null) {
 			securityContext.withSSLContext(sslInputs.getSslContext(), sslInputs.getTrustManager());
@@ -160,7 +160,7 @@ public class DatabaseClientPropertySource {
 		return securityContext;
 	}
 
-	private DatabaseClientFactory.SecurityContext newSecurityContext(String type, SSLInputs sslInputs) {
+	private DatabaseClientFactory.SecurityContext newSecurityContext(String type, SSLUtil.SSLInputs sslInputs) {
 		switch (type.toLowerCase()) {
 			case DatabaseClientBuilder.AUTH_TYPE_BASIC:
 				return newBasicAuthContext();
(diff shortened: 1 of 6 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
```diff
@@ -271,28 +278,34 @@ private DatabaseClientFactory.SSLHostnameVerifier determineHostnameVerifier() {
...
 		SSLContext sslContext = getSSLContext();
 		if (sslContext != null) {
-			return new SSLInputs(sslContext, userTrustManager);
+			return new SSLUtil.SSLInputs(sslContext, userTrustManager);
 		}
 
-		// Approaches 2 and 3 - user defines an SSL protocol.
-		// Approach 2 - "default" is a convenience for using the JVM's default SSLContext.
-		// Approach 3 - create a new SSLContext, and initialize it if the user-provided TrustManager is not null.
+		// Approach 2 - user wants two-way SSL via a keystore.
```
rjrudin, line 290: We had 4 approaches before, we now have 5 - the new approach number 2 is for two-way SSL via a keystore.

marklogic-client-api/src/main/java/com/marklogic/client/impl/SSLUtil.java
```diff
@@ -40,29 +51,140 @@ static X509TrustManager getDefaultTrustManager() {
...
+			throw new RuntimeException("No trust managers found using algorithm: " + trustManagerAlgorithm);
 		}
 		if (!(trustManagers[0] instanceof X509TrustManager)) {
 			throw new RuntimeException("Default trust manager is not an X509TrustManager: " + trustManagers[0]);
 		}
 		return trustManagers;
 	}
+
+	/**
+	 * Captures the oft-repeated boilerplate Java code for creating an SSLContext based on a key store.
```
rjrudin, line 93: This is the location of the main annoying code that a user would otherwise have to write themselves.

marklogic-client-api/src/test/java/com/marklogic/client/test/ssl/TwoWaySSLTest.java
```diff
@@ -96,15 +90,19 @@ public static void teardown() {
...
 	@Test
-	void digestAuthentication() throws Exception {
+	void digestAuthentication() {
 		if (Common.USE_REVERSE_PROXY_SERVER) {
 			return;
 		}
 
 		// This client uses our Java KeyStore file with a client certificate in it, so it should work.
 		DatabaseClient clientWithCert = Common.newClientBuilder()
+			.withKeyStorePath(keyStoreFile.getAbsolutePath())
```
rjrudin, line 100: This is the main value of this story - the user only needs to provide a couple properties for 2-way SSL now instead of all the annoying code for building their own SSL context with all the proper error handling.

anu3990 (approved):

PR #1615, opened 2023-10-24
DEVEXP-563 Added support for other 11.1 Optic methods

rjrudin:
Was able to code-generate for the document format/root/permission queries, but DBQ-643 in the server prevents them from working. 

joinDocAndUri appears to be good-to-go.

I had to manually add the string constructors for cts.point and cts.polygon as the code generator does not yet generate those. Added comments to capture that fact. 

Finally, renamed the two XML TDEs that were using .tdex as an extension - .xml is preferred.


BillFarber (approved):
LGTM - but I'm flying blind.

PR #1622, opened 2023-11-29
DEVEXP-650 Fixing search that includes metadata values

rjrudin:
Need to use "toString()" instead of "name()" so that "metadatavalues" is converted into "metadata-values". This works fine for reading documents but not for searching documents.

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/StringSearchTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/StringSearchTest.java
@@ -20,6 +20,10 @@ import com.marklogic.client.ForbiddenUserException;
 import com.marklogic.client.ResourceNotFoundException;
 import com.marklogic.client.ResourceNotResendableException;
 import com.marklogic.client.admin.QueryOptionsManager;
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/StringSearchTest.java
```diff
@@ -54,9 +58,24 @@ public static void beforeClass() {
...
+	  mgr.write("/metadata/test.xml",
+		  new DocumentMetadataHandle().withMetadataValue("hello", "world"),
+		  new StringHandle("<test>metadataabc</test>"));
+
+	  mgr.setMetadataCategories(DocumentManager.Metadata.METADATAVALUES);
+	  DocumentPage page = mgr.search(Common.client.newQueryManager().newStructuredQueryBuilder().term("metadataabc"), 1);
+	  while (page.hasNext()) {
+		  DocumentRecord record = page.next();
+		  assertEquals("/metadata/test.xml", record.getUri());
+	  }
```
BillFarber, line 73: Would it be worth asserting that the metadataValue was returned?

BillFarber (approved):

Issue #1640, opened 2024-02-21
search:and-not-query is ignored in QueryBatcher


marcopacurariu3:
### Version of MarkLogic Java Client API
6.5.0
### Version of MarkLogic Server
11.0.2
### Java version
JDK 17
### OS and version
ProductName:		macOS
ProductVersion:		13.5
BuildVersion:		22G74

### Input: Some code to illustrate the problem

```
final RawCombinedQueryDefinition structQueryDef = queryManager
            .newRawCombinedQueryDefinition(new StringHandle(
                """
                    <search:search xmlns:search="http://marklogic.com/appservices/search">
                       <search:query>
                          <query xmlns="http://marklogic.com/appservices/search" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
                             <search:and-query>
                                   <search:range-constraint-query>
                                      <search:constraint-name>application</search:constraint-name>
                                      <search:value>testapp</search:value>
                                      <search:range-operator>EQ</search:range-operator>
                                   </search:range-constraint-query>
                                   <search:and-not-query>
                                      <search:annotation type="searchable-collections" />
                                      <search:positive>
                                         <search:collection-query>
                                            <search:uri>workspace/testapp/someCollection</search:uri>
                                         </search:collection-query>
                                      </search:positive>
                                      <search:negative>
                                            <search:collection-query>
                                               <search:uri>workspace/testapp/ignoredCollection</search:uri>
                                            </search:collection-query>
                                      </search:negative>
                                   </search:and-not-query>
                             </search:and-query>
                          </query>
                       </search:query>
                    </search:search>
                    """), "all");

```

While executing a query that contains a _search:and-not-query_, I noticed that the QueryBatcher is simply ignoring that part. However, the same query that is executed for a simple search does work. 

` final SearchHandle result = queryManager.search(structQueryDef, new SearchHandle());
        result.getMatchResults();`
        
The above search works. Same query passed to a QueryBatcher and calling _batcher.getItems()_ does not take into account the _and-not-query_ part.

### Actual output: What did you observe? What errors did you see? Can you attach the logs? (Java logs, MarkLogic logs)
Executing the above query does not take into account the _and-not-query_ for QueryBatcher.

### Expected output: What specifically did you expect to happen?
The _and-not-query_ is taken into account for QueryBatcher.

### Alternatives: What else have you tried, actual/expected?
[...]

PR #1641, opened 2024-02-21
Added test for and-not query

rjrudin:
Associated to #1640, trying to reproduce the error.

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/StructuredSearchTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/StructuredSearchTest.java
@@ -15,13 +15,17 @@
  */
 package com.marklogic.client.test;
 
+import com.marklogic.client.datamovement.DataMovementManager;
+import com.marklogic.client.datamovement.QueryBatcher;
+import com.marklogic.client.document.DocumentWriteSet;
 import com.marklogic.client.document.GenericDocumentManager;
+import com.marklogic.client.document.XMLDocumentManager;
+import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.Format;
 import com.marklogic.client.io.SearchHandle;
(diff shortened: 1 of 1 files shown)


marklogic-client-api/src/test/java/com/marklogic/client/test/StructuredSearchTest.java
```diff
@@ -15,168 +15,240 @@
...
+			"            <uri>test1</uri>\n" +
+			"          </collection-query>\n" +
+			"        </positive-query>\n" +
+			"        <negative-query>\n" +
+			"          <collection-query>\n" +
+			"            <uri>test2</uri>\n" +
+			"          </collection-query>\n" +
+			"        </negative-query>\n" +
+			"      </and-not-query>\n" +
+			"    </and-query>\n" +
```
hansenmc, line 232: Wouldn't we expect an and-not-query to be:


			"      <and-not-query>\n" +
			"        <positive-query>\n" +
            "          <and-query>\n" +
			"            <term-query>\n" +
			"              <text>world</text>\n" +
			"            </term-query>\n" +
			"            <collection-query>\n" +
			"              <uri>test1</uri>\n" +
			"            </collection-query>\n" +
			"          </and-query>\n" +
			"        </positive-query>\n" +
			"        <negative-query>\n" +
			"          <collection-query>\n" +
			"            <uri>test2</uri>\n" +
			"          </collection-query>\n" +
			"        </negative-query>\n" +
			"      </and-not-query>\n" +

rjrudin: It could be like that, I was just modeling it after the query posted in #1640 . I think the details of the query don't matter (unless it's a specific query that is not working in conjunction with an and-not-query), just need to verify that the and-not-query is being honored in a combined query with and without DMSDK.

PR #1647, opened 2024-03-21
Adding back orderBy to some Optic tests

rjrudin:
These were removed as part of MLE-805 (formerly bug 60065) when these tests started failing on 11-nightly due to the orderBy. I added orderBy back and ran them on 11.1.0, 11-nightly, and 12-nightly, and they all passed.


BillFarber (approved):

rjrudin:
Merging this, as the one failing test is also failing on the develop branch, will fix that separately.

PR #1656, opened 2024-04-08
Fixes for regression failures on 12-nightly

rjrudin:
Also removed unused "CONTRIBUTING" file - we're using the one in the root directory.


anu3990 (approved):

rjrudin:
The test failure is an intermittent DMSDK failure, so ignoring.

PR #1661, opened 2024-05-01
Fixing bug where a 'naked' properties fragment cannot be written

rjrudin:
The check on content being null was overly restrictive, as v1/documents allows for null content. This may cause some other DMSDK test to fail, in which case that test will be modified as null content is allowable here.


BillFarber (approved):
pending test fix

BillFarber (approved):

PR #1672, opened 2024-07-03
Can now use search options in ML12 for BM25, Zero, and Random.

BillFarber:
Also adds a setting for BM25 Length Weight.

Added a couple of tests. These tests are of limited use here, but they do verify that "bm25" and "zero" are acceptable scoring options.

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/type/PlanSearchOptions.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/type/PlanSearchOptions.java
@@ -15,19 +15,34 @@
  */
 package com.marklogic.client.type;
 
-// IMPORTANT: Do not edit. This file is generated.
-
 /**
  * An option controlling the scoring and weighting of fromSearch()
  * for a row pipeline.
  */
 public interface PlanSearchOptions {
-    XsDoubleVal getQualityWeight();
+	/**
+	 * Changed in release 6.7.0 to return a float, as the server requires a float and throws an error on a double.
+	 */
+	XsFloatVal getQualityWeight();
     ScoreMethod getScoreMethod();
-    PlanSearchOptions withQualityWeight(double qualityWeight);
-    PlanSearchOptions withQualityWeight(XsDoubleVal qualityWeight);
+	/**
+	 * @since 6.7.0
+	 */
+	XsDoubleVal getBm25LengthWeight();
+	/**
+	 * Changed in release 6.7.0 to return a float, as the server requires a float and throws an error on a double.
+	 */
+    PlanSearchOptions withQualityWeight(float qualityWeight);
+	/**
+	 * Changed in release 6.7.0 to return a float, as the server requires a float and throws an error on a double.
+	 */
+    PlanSearchOptions withQualityWeight(XsFloatVal qualityWeight);
     PlanSearchOptions withScoreMethod(ScoreMethod scoreMethod);
+	/**
+	 * @since 6.7.0
+	 */
+	PlanSearchOptions withBm25LengthWeight(double bm25LengthWeight);
     enum ScoreMethod {
-        LOGTFIDF, LOGTF, SIMPLE;
+        LOGTFIDF, LOGTF, SIMPLE, BM25, ZERO, RANDOM;
     }
 }
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/rows/FromSearchWithOptionsTest.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/rows/FromSearchWithOptionsTest.java
@@ -1,0 +1,73 @@
+package com.marklogic.client.test.rows;
+
+import com.marklogic.client.FailedRequestException;
+import com.marklogic.client.expression.PlanBuilder;
+import com.marklogic.client.row.RowRecord;
+import com.marklogic.client.test.junit5.RequiresML12;
+import com.marklogic.client.type.PlanSearchOptions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith(RequiresML12.class)
+class FromSearchWithOptionsTest extends AbstractOpticUpdateTest {
+
+	@Test
+	void bm25() {
+//		Note that this does not actually test that the scoring is correct.
+//		It only tests that including the BM25 scoring option and a valid bm25LengthWeight do not cause any problems.
+		rowManager.withUpdate(false);
+		PlanSearchOptions options = op.searchOptions()
+			.withScoreMethod(PlanSearchOptions.ScoreMethod.BM25)
+			.withBm25LengthWeight(0.25);
+		List<RowRecord> rows = resultRows(
+			op.fromSearch(op.cts.wordQuery("contents"), null, null, options)
+				.offsetLimit(0, 5)
+		);
+		assertEquals(5, rows.size());
+	}
+
+	@Test
+	void badBm25LengthWeight() {
+		rowManager.withUpdate(false);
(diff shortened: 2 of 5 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/type/PlanSearchOptions.java
```diff
@@ -15,7 +15,8 @@
  */
 package com.marklogic.client.type;
 
-// IMPORTANT: Do not edit. This file is generated.
+// (THIS COMMENT IS WRONG) IMPORTANT: Do not edit. This file is generated.
```
rjrudin, line 18: Just remove this comment since it's wrong.

marklogic-client-api/src/test/java/com/marklogic/client/impl/OpticPlanSearchOptionsTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.Iterator;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith(RequiresML12.class)
+class OpticPlanSearchOptionsTest {
```
rjrudin, line 21: I'd like this to be similar to the existing `NewOpticMethodsInElevenDotOneTest`, which is in the `test.rows` package as well with all the other Optic tests. 

Are all of these only in 12 though - none of these new features are in 11.3? If so, still put this into `test.rows` and call it `FromSearchWithOptionsTest` as we'll probably have too many new things in 12 to put them into one generic test. 

Also, extend `AbstractOpticUpdateTest`, that provides a number of helpers to make the test easier to write.

marklogic-client-api/src/test/java/com/marklogic/client/impl/OpticPlanSearchOptionsTest.java
```diff
@@ -0,0 +1,53 @@
...
+	@BeforeEach
+	void beforeEach() {
+		client = Common.newClient();
+		rowManager = client.newRowManager();
+		op = rowManager.newPlanBuilder();
+	}
+
+	@Test
+	void bm25() {
+		PlanBuilderBaseImpl.PlanSearchOptionsImpl options =
```
rjrudin, line 35: We don't want users to have to instantiate these classes themselves (generally, we don't want a user interacting with any class with "Impl" in the name). There's a `PlanBuilderBase.searchOptions()` method (that is unfortunately not used by any tests yet) that the user should call instead.

marklogic-client-api/src/test/java/com/marklogic/client/impl/OpticPlanSearchOptionsTest.java
```diff
@@ -0,0 +1,53 @@
...
+			new PlanBuilderBaseImpl.PlanSearchOptionsImpl(new PlanBuilderSubImpl(), null, PlanSearchOptions.ScoreMethod.ZERO, null);
+		PlanBuilder.ModifyPlan plan = op.fromSearch(op.cts.wordQuery("contents"), null, null, options).offsetLimit(0, 5);;
+		RowSet<RowRecord> rowSet = rowManager.resultRows(plan);
```
rjrudin, line 38: Use `resultRows` in `AbstractOpticUpdateTest`, which also returns a far more useful List instead of a RowSet.

rjrudin (changes requested):

marklogic-client-api/src/test/java/com/marklogic/client/impl/OpticPlanSearchOptionsTest.java
BillFarber: According to Aaron, these changes are only being made on the "develop" and "release-12.0.0-ea1" branches, which means they are only in MarkLogic 12+.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/FromSearchWithOptionsTest.java
```diff
@@ -0,0 +1,40 @@
...
+	@Test
+	void bm25() {
+		PlanSearchOptions options = op.searchOptions()
+			.withScoreMethod(PlanSearchOptions.ScoreMethod.BM25)
+			.withBm25LengthWeight(20.5);
+		List<RowRecord> rows = resultRows(
+			op.fromSearch(op.cts.wordQuery("contents"), null, null, options)
+				.offsetLimit(0, 5)
+		);
+		assertEquals(5, rows.size());
```
rjrudin, line 25: Include a message here noting that we're not actually testing that BM25 scoring is being calculated correctly - we trust the server to do that. We just want to make sure that selecting BM25 and including a length/weight doesn't cause any problems. 

Ideally though, there'd be some way to detect that the length/weight is actually submitted to the server - for all we know, the Java Client may be ignoring it. To achieve that - the best approach is probably to try putting in an invalid value - maybe a negative one, or something outside the accepted bounds, if such bounds exist? - and then verifying that the server throws an error. 

If that's not possible, then just include the assertion message and merge away.

rjrudin (approved):
Approved, but see if it's possible to make a test that fails on an invalid BM25 length weight.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/FromSearchWithOptionsTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+	@Test
+	void badBm25LengthWeight() {
+		rowManager.withUpdate(false);
+		PlanSearchOptions options = op.searchOptions()
+			.withScoreMethod(PlanSearchOptions.ScoreMethod.BM25)
+			.withBm25LengthWeight(99);
+		PlanBuilder.ModifyPlan plan = op.fromSearch(op.cts.wordQuery("contents"), null, null, options)
+			.offsetLimit(0, 5);
+		Exception exception = assertThrows(FailedRequestException.class, () -> {
```
rjrudin, line 41: No need for braces, which allows this to be on one line too.

rjrudin (approved):

marklogic-client-api/src/main/java/com/marklogic/client/type/PlanSearchOptions.java
```diff
@@ -15,19 +15,19 @@
...
 
-// IMPORTANT: Do not edit. This file is generated.
-
 /**
  * An option controlling the scoring and weighting of fromSearch()
  * for a row pipeline.
  */
 public interface PlanSearchOptions {
-    XsDoubleVal getQualityWeight();
+	XsFloatVal getQualityWeight();
```
rjrudin, line 23: Given that this is changing due to a bug and it's a public API, add the following javadoc for it:

```
/**
 * Changed in release 6.7.0 to return a float, as the server requires a float and throws an error on a double. 
 */ 
```

marklogic-client-api/src/main/java/com/marklogic/client/type/PlanSearchOptions.java
```diff
@@ -15,19 +15,19 @@
...
     ScoreMethod getScoreMethod();
-    PlanSearchOptions withQualityWeight(double qualityWeight);
-    PlanSearchOptions withQualityWeight(XsDoubleVal qualityWeight);
+	XsDoubleVal getBm25LengthWeight();
```
rjrudin, line 25: For each new method in this, add the following javadoc:

```
/**
 * @since 6.7.0
 */
```

We haven't been good with that in the Optic plan builder stuff, but it's mostly been added everywhere else in the Java Client since 6.0.

rjrudin (approved):
LG, just a few bookkeeping javadocs to add.

rjrudin (approved):

PR #1673, opened 2024-07-08
Can now use search options with op.fromSearchDocs.

BillFarber:


marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
```diff
@@ -54,12 +54,25 @@ public interface PlanBuilderBase {
      * @return  a ModifyPlan object
      */
     PlanBuilder.AccessPlan fromSearchDocs(CtsQueryExpr query, String qualifierName);
+	/**
+	 * Provides a convenience for matching documents and constructing rows with the score,
+	 * document uri, and document content. The convenience is equivalent to chaining
```
rjrudin, line 59: Not urgent, but if you feel up for it, you could try changing whatever the source of this comment is to use "URI" instead of "uri" (pet peeve of mine to see "uri" unless it's referring to the column name).

rjrudin (approved):

marklogic-client-api/src/main/java/com/marklogic/client/expression/PlanBuilderBase.java
BillFarber: This source is not generated, so I've updated it directly.

PR #1674, opened 2024-07-09
Optic Generation updates for release-12.0.0-ea1

BillFarber:


rjrudin (approved):
LG, though tests are needed for at least one of the new vector operations, which can be either here or in a follow-up PR.

PR #1677, opened 2024-07-10
Added tests for the new Optic vector functions.

BillFarber:


marklogic-client-api/src/test/java/com/marklogic/client/test/rows/VectorTest.java
```diff
@@ -0,0 +1,155 @@
...
+				.bind(op.as("dimension", op.vec.dimension(op.col("sampleVector"))))
+				.bind(op.as("normalize", op.vec.normalize(op.col("sampleVector"))))
+				.bind(op.as("magnitude", op.vec.magnitude(op.col("sampleVector"))))
+				.bind(op.as("get", op.vec.get(op.col("sampleVector"), op.xs.integer(2))))
+				.bind(op.as("add", op.vec.add(op.col("embedding"),op.col("sampleVector"))))
+				.bind(op.as("subtract", op.vec.subtract(op.col("embedding"),op.col("sampleVector"))))
+				.bind(op.as("base64Encode", op.vec.base64Encode(op.col("sampleVector"))))
+				.bind(op.as("base64Decode", op.vec.base64Decode(op.col("base64Encode"))))
+				.bind(op.as("subVector", op.vec.subvector(op.col("sampleVector"), op.xs.integer(1), op.xs.integer(1))))
+//				There seems to be a server bug preventing vectorScore from working. See ticket MLE-15508
```
rjrudin, line 130: Make this a "@Disabled" test instead (with just the stuff in it that doesn't work). That's much easier to find later vs commented out code. Arguably, we can even let it fail, but I think "Disabled" is good.

marklogic-client-api/src/test/java/com/marklogic/client/test/rows/VectorTest.java
```diff
@@ -0,0 +1,155 @@
...
+				.bind(op.as("sampleVector", sampleVector))
+				.bind(op.as("summaryCosineSim", op.vec.cosineSimilarity(op.col("embedding"),op.col("sampleVector"))))
+				.select(op.col("name"), op.col("summary"), op.col("summaryCosineSim"))
+          		.orderBy(op.desc(op.col("summaryCosineSim")))
+				.limit(5);
+		List<RowRecord> rows = resultRows(plan);
+		assertEquals(2, rows.size());
+		rows.forEach(row -> {
+			// Simply a sanity check that the Cosine Similarity score is valid.
+			Assertions.assertTrue(
```
rjrudin, line 45: For all the "assertTrue" assertions - without a message, if any of these fail, it's usually impossible to know what the actual value is; JUnit won't show anything. And with this being brand new stuff - and thus more likely to change - I think we'll save time in the future by improving these assertions to include a message along the lines of "Unexpected value: " + something - i.e.

```
double val = row.getDouble("summaryCosineSim");
assertTrue(val > 0 && val < 1, "Unexpected value: " + value);
```

Note that you don't need "Assertions." either since you've already statically imported assertTrue.

rjrudin (changes requested):

rjrudin (approved):

PR #1682, opened 2024-07-26
Created a test for a new bug.

BillFarber:


marklogic-client-api/src/test/java/com/marklogic/client/test/rows/VectorTest.java
```diff
@@ -108,4 +109,23 @@ void cosineSimilarity_InvalidVector() {
...
+			op.fromSearchDocs(
+				op.cts.andQuery(
+					op.cts.documentQuery("/optic/vectors/alice.json"),
+					op.cts.elementQuery(
+						"person",
+						op.cts.andQuery()
+    				)
+  				))
+			.bind(op.as("embedding", op.vec.vector(op.xpath("doc", "/person/embedding"))))
+			.limit(5);
```
rjrudin, line 127: Remove the limit call here, it's implying that it's needed but since the assertion expects a single document, it must not be.

rjrudin (approved):

PR #1701, opened 2024-09-03
Auto-gen for shortest path.

anu3990:
* OS and version


rjrudin (changes requested):
Several issues here:

1. This PR introduces compilation errors by removing code that we need.
2. It doesn't add anything pertaining to the shortest path feature.
3. It changes file permissions on files that we don't want changed. 
4. It changes the copyright to a non-standard one on a couple dozen files. 

We know from our experience with the patch feature that the code generator is not reliable, so likely will need to add support manually for the shortest path function.

PR #1702, opened 2024-09-11
MLE-16370 : Run optic auto-generation for op:shortest-path

anu3990:
* OS and version


rjrudin (changes requested):
Need to revert all the copyright changes - ideally, try fixing the code generator to produce the new standard copyright instead of the old one. 

We also need a test for this - we know from our experience back in the spring that the generated code isn't guaranteed to work. So add a `ShortestPathTest` in the com.marklogic.client.test.rows package and add the annotation so that it only runs on MarkLogic 12.

Also - for the new shortestPath methods in the interface - add `@since 7.1.0` annotations to the Javadoc comments so we have an easy way of knowing when they were added. In fact, make it `@since 7.1.0; requires MarkLogic 12`.

PR #1705, opened 2024-10-07
Can now specify a page length of 0 when searching with OkHttpServices.

BillFarber:
I think I'm covering the ways that this change might get called, though I'm not certain.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ ml-development-tools/src/test/ml-modules
 ml-development-tools/src/test/java/com/marklogic/client/test/dbfunction/generated
 
 .vscode
+docker-compose.yaml
+docker/
--- a/marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/search/SearchWithPageLengthTest.java
+++ b/marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/search/SearchWithPageLengthTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright © 2024 MarkLogic Corporation. All Rights Reserved.
+ */
+
+package com.marklogic.client.fastfunctest.search;
+
+import com.marklogic.client.fastfunctest.AbstractFunctionalTest;
+import com.marklogic.client.io.*;
+import com.marklogic.client.query.*;
+import org.junit.jupiter.api.*;
+
+import java.io.FileNotFoundException;
+
+import static org.custommonkey.xmlunit.XMLAssert.*;
+
+
+class SearchWithPageLengthTest extends AbstractFunctionalTest {
+
+	static final QueryManager queryMgr = client.newQueryManager();
+	StringQueryDefinition qd = queryMgr.newStringDefinition();
+
+	@BeforeEach
+	public void testSetup() throws FileNotFoundException {
+		String[] filenames = {"constraint1.xml", "constraint2.xml", "constraint3.xml", "constraint4.xml", "constraint5.xml"};
+		for (String filename : filenames) {
+			writeDocumentUsingInputStreamHandle(client, filename, "/return-results-false/", "XML");
+		}
+	}
+
+	@AfterEach
+	public void testCleanUp() {
+		deleteDocuments(connectAsAdmin());
+	}
+
+	@Test
+	void testSearchPageLengthZero()  {
+		queryMgr.setPageLength(0);
+		qd.setCriteria("Bush");
+		SearchHandle resultsHandle = new SearchHandle();
+		queryMgr.search(qd, resultsHandle);
+
+		assertEquals(0, resultsHandle.getPageLength());
+		assertEquals("The server allows for a zero page length, which results in no results being returned.",
+			0, resultsHandle.getMatchResults().length);
+	}
+
+
+	@Test
+	void testSearchPageLengthNegative() {
+		queryMgr.setPageLength(-1);
+		qd.setCriteria("Bush");
+		SearchHandle resultsHandle = new SearchHandle();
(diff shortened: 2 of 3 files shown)


marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/TestSearchPageLength.java
```diff
@@ -0,0 +1,73 @@
...
+		qd.setCriteria("Foo");
+		DOMHandle resultsHandle = new DOMHandle();
+		queryMgr.search(qd, resultsHandle);
+
+		Document resultDoc = resultsHandle.get();
+		assertXpathEvaluatesTo("10", "string(//*[local-name()='response']//@*[local-name()='page-length'])", resultDoc);
+	}
+
+	@Test
+	void testOkHttpPageLengthParamaterZero() throws XpathException {
```
rjrudin, line 47: We don't need this test or the one below it - this isn't a public API for our users. Only need to test `queryManager.search`.

test-app/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
+name: marklogic-javaclient-test-app
```
rjrudin, line 1: Hmm - add a note to this for now saying that it's for optional usage. I'm reluctant to add this without also 1) Updating the CONTRIBUTING guide to recommend Docker usage, and 2) Updating the Jenkinsfile to use Docker as well. In fact - for those reasons, don't include this yet. But getting a JIRA ticket and PR together to take care of those two other things would be a good step forward for this repository.

marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/TestSearchPageLength.java
```diff
@@ -0,0 +1,73 @@
...
+import com.marklogic.client.io.*;
+import com.marklogic.client.query.*;
+import org.custommonkey.xmlunit.exceptions.XpathException;
+import org.junit.jupiter.api.*;
+import org.w3c.dom.Document;
+
+import static org.custommonkey.xmlunit.XMLAssert.*;
+
+
+class TestSearchPageLength extends AbstractFunctionalTest {
```
rjrudin, line 18: I have no idea why so many tests start with "Test" instead of ending with "Test", when the latter is the very common convention for JUnit. Having everything start with "Test" is also annoying for finding test classes via meaningful verbs at the start of the test class. So do `SearchWithPageLengthTest` to make it obvious what this test class is covering. 

Also, you can see that not much thought was given to packaging structures - i.e. there's a ton of tests in one package. We can start improving that though - so do `com.marklogic.client.fastfunctest.search` for the package.

marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/TestSearchPageLength.java
```diff
@@ -0,0 +1,73 @@
...
+
+	static final QueryManager queryMgr = client.newQueryManager();
+	StringQueryDefinition qd = queryMgr.newStringDefinition();
+
+	@Test
+	void testSearchPageLengthZero() throws XpathException {
+		queryMgr.setPageLength(0);
+		qd.setCriteria("Foo");
+		DOMHandle resultsHandle = new DOMHandle();
```
rjrudin, line 27: Use a SearchHandle instead - that avoids awkward XPath expressions. And it lets you add a nice assertion message - i.e. "The server allows for a zero page length, which results in no results being returned." And also assert that no results are actually returned - i.e. `assertEquals(0, handle.getMatchResults().length)` (I'm curious if getMatchResults() returns null there or not).

marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/TestSearchPageLength.java
```diff
@@ -0,0 +1,73 @@
...
+
+	@Test
+	void testSearchPageLengthNegative() throws XpathException {
+		queryMgr.setPageLength(-1);
+		qd.setCriteria("Foo");
+		DOMHandle resultsHandle = new DOMHandle();
+		queryMgr.search(qd, resultsHandle);
+
+		Document resultDoc = resultsHandle.get();
+		assertXpathEvaluatesTo("10", "string(//*[local-name()='response']//@*[local-name()='page-length'])", resultDoc);
```
rjrudin, line 43: This is surprising behavior by the server - i.e. it appears to be ignoring a negative value and using the server default. That is worth capturing in an assertion message - i.e. that we don't expect an error from the server, the server just ignores the input.

test-app/docker-compose.yaml
BillFarber: https://progresssoftware.atlassian.net/browse/MLE-17231

marklogic-client-api-functionaltests/src/test/java/com/marklogic/client/fastfunctest/TestSearchPageLength.java
BillFarber: The server isn't ignoring the negative value, it isn't getting passed.
Why would you have a negative page length?

rjrudin: Oh right - that's what the assertion message should state then - i.e. that a negative value isn't sent (instead of an error being thrown), so we expect the server default page length of 10 to be used.

rjrudin (approved):

PR #1707, opened 2024-10-09
Converting Jenkinsfile to use a docker-compose file

BillFarber:
Let's see if this first step works, and then I'll try more.

Just a draft PR for now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,17 @@ and add the following (you can override additional properties as necessary):
 Note that additional properties are defined via `./tests-app/gradle.properties`, though it is not expected that these
 properties will need to be changed.
 
-The application is then deployed via the following command:
+The tests require a MarkLogic instance with access to several ports (8000,8001,8002,8012,8014,8020). That instance may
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -28,7 +28,17 @@ and add the following (you can override additional properties as necessary):
...
 properties will need to be changed.
 
-The application is then deployed via the following command:
+The tests require a MarkLogic instance with access to several ports (8000,8001,8002,8012,8014,8020). That instance may
+be a local instance or it may be running in a Docker container. If you would like to create a Docker container with the
+instance, you may create the container with the following commands (starting in the project root directory):
+
+```
+cd test-app
+docker-compose up -d --build
```
rjrudin, line 37: `docker compose` as that's the new way to do it.

rjrudin (approved):
One small tweak - and then either in this PR or a subsequent one, I'm hoping the noDocker / RPM support can be dumped by changing "MARKLOGIC_TAG" to be the entire image name. And we can get rid of 10.0-10.2 if there's not a tag for it. latest-10 is sufficient.

PR #1709, opened 2024-10-11
Converting Jenkinsfile to use a docker-compose file

BillFarber:
Switch to images and artifactory.

Take 3 - Trying to run the Jenkinsfile multiple times.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,31 +1,39 @@
 @Library('shared-libraries') _
 
 def getJava(){
-    if(env.JAVA_VERSION=="JAVA17"){
-        return "/home/builder/java/jdk-17.0.2"
-    }else if(env.JAVA_VERSION=="JAVA11"){
-        return "/home/builder/java/jdk-11.0.2"
-    }else if(env.JAVA_VERSION=="JAVA21"){
+	if(env.JAVA_VERSION=="JAVA17"){
+		return "/home/builder/java/jdk-17.0.2"
+	}else if(env.JAVA_VERSION=="JAVA11"){
+		return "/home/builder/java/jdk-11.0.2"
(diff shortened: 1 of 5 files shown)


test-app/.env
```diff
@@ -0,0 +1,3 @@
+# Defines environment variables for docker-compose.
+# Can be overridden via e.g. `MARKLOGIC_TAG=latest-10.0 docker-compose up -d --build`.
+MARKLOGIC_IMAGE=progressofficial/marklogic-db:latest
```
rjrudin, line 3: Based on the problems we had with this tag in the Jenkinsfile for the Spark connector - pinging @SameeraPriyathamTadikonda  to confirm - I think "11.3.0-ubi" should be the default instead of "latest".

Jenkinsfile
```diff
@@ -1,31 +1,39 @@
...
+}
 
-            sh label:'deploy test app', script: '''#!/bin/bash
-                export JAVA_HOME=$JAVA_HOME_DIR
-                export GRADLE_USER_HOME=$WORKSPACE/$GRADLE_DIR
-                export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
-                cd java-client-api
-                ./gradlew -i mlDeploy mlReloadSchemas -PmlForestDataDirectory=/space
-            '''
+def runAllTests(Boolean useReverseProxy, String image){
```
rjrudin, line 35: Not important now, but a nice refactor (which would have been useful before this PR too) would to make two separate functions:

- `runTests(image)`
- `runTestsWithReverseProxy(image)`

That gets rid of any confusion below over what the boolean refers to, and what position it should be in the arg list. 

Again, not important now, but a good refactor after this PR.

rjrudin (approved):

PR #1720, opened 2024-11-05
Added test for the server-side vec:vector function fix.

BillFarber:


marklogic-client-api/src/test/java/com/marklogic/client/test/rows/VectorTest.java
```diff
@@ -126,4 +126,14 @@ void bindVectorFromDocs() {
 		List<RowRecord> rows = resultRows(plan);
 		assertEquals(1, rows.size());
 	}
+
+	@Test
+	void vecVectorWithCol() {
+		PlanBuilder.ModifyPlan plan =
```
rjrudin, line 132: Need to do a DSL script here - we already have tests that verify that the builder works.  ColumnInfoTest has a couple examples of using a DSL script - e.g. 

```
String query = "op.fromView('javaClient', 'allTypes')";
		RawQueryDSLPlan plan = rowManager.newRawQueryDSLPlan(new StringHandle(query));
```

rjrudin (changes requested):

rjrudin (approved):

PR #1743, opened 2025-04-02
Can now specify SSL Protocol when creating a CertificateAuthContext.

BillFarber:
I'm not sure what else, if anything, needs to be done for https://progresssoftware.atlassian.net/browse/MLE-19397

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
@@ -913,8 +913,9 @@ public class DatabaseClientFactory {
   }
 
   public static class CertificateAuthContext extends AuthContext {
-    String certFile;
-    String certPassword;
+	  private static final String DEFAULT_SSL_PROTOCOL = "TLSv1.2";
+	  String certFile;
+	  String certPassword;
 
 	/**
      * Creates a CertificateAuthContext by initializing the SSLContext of the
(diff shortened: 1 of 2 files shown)


marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -973,13 +974,42 @@ public CertificateAuthContext(SSLContext context, SSLHostnameVerifier verifier,
      *           UnrecoverableKeyException.
      */
     public CertificateAuthContext(String certFile, X509TrustManager trustManager)
-        throws CertificateException, IOException,
-        UnrecoverableKeyException, KeyManagementException {
+        	throws CertificateException, IOException, UnrecoverableKeyException, KeyManagementException {
```
rjrudin, line 977: Have this constructor call the next one, passing in TLSv1.2 as the value of protocol. Same goes for the other pair of constructors.

rjrudin (changes requested):
The other step is to make DatabaseClientPropertySource use the value of sslProtocol when it creates a CertificateAuthContext. newCertificateAuthContext will need that value so it can pass it into the constructor.

marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -1037,7 +1098,7 @@ private SSLContext createSSLContext()
         }
         keyManagerFactory.init(keyStore, certPassword.toCharArray());
         keyMgr = keyManagerFactory.getKeyManagers();
-        sslContext = SSLContext.getInstance("TLSv1.2");
+        sslContext = SSLContext.getInstance(sslProtocol);
```
rjrudin, line 1101: For safety reasons, I think this should say "if protocol is null, then use the default value".

rjrudin: The null check needs to be there though. The DatabaseClientPropertySource is just one way of building a CertificateAuthContext. A user can build one themselves.

marklogic-client-api/src/main/java/com/marklogic/client/DatabaseClientFactory.java
```diff
@@ -973,13 +974,42 @@ public CertificateAuthContext(SSLContext context, SSLHostnameVerifier verifier,
...
     public CertificateAuthContext(String certFile, X509TrustManager trustManager)
-        throws CertificateException, IOException,
-        UnrecoverableKeyException, KeyManagementException {
+        	throws CertificateException, IOException, UnrecoverableKeyException, KeyManagementException {
         this.certFile = certFile;
         this.trustManager = trustManager;
         this.certPassword = "";
         this.sslContext = createSSLContext();
-      }
+	}
```
rjrudin, line 982: Have the above constructor invoke the next one - e.g. `this(certFile, trustManager, certPassword, null)` - and the next one will apply the logic of defaulting a null password to `""` and defaulting the protocol value if null.

rjrudin (changes requested):

marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
```diff
@@ -159,7 +161,11 @@ private DatabaseClientFactory.SecurityContext newSecurityContext() {
 		final String authType = determineAuthType(connectionString);
 
 		final SSLUtil.SSLInputs sslInputs = buildSSLInputs(authType);
-		DatabaseClientFactory.SecurityContext securityContext = newSecurityContext(authType, connectionString, sslInputs);
+		String sslProtocol = getSSLProtocol(authType);
+		if ("default".equalsIgnoreCase(sslProtocol) || sslProtocol == null) {
```
rjrudin, line 165: You don't need this here, as it's only for cert auth and the cert auth will handle the default value. 

In fact, to make this more clear, rename the sslProtocol variable to `sslProcotolForCertificateAuth`. That's making it clear that cert auth is the oddball here that can't be happy with just an SSLContext; it needs an SSL protocol too so it can build its own SSLContext based on a cert file/password.

rjrudin (changes requested):
Sorry, one more tweak for clarity.

marklogic-client-api/src/main/java/com/marklogic/client/impl/DatabaseClientPropertySource.java
```diff
@@ -29,6 +29,8 @@ public class DatabaseClientPropertySource {
 
 	private static final String PREFIX = DatabaseClientBuilder.PREFIX;
 
+	private static final String DEFAULT_SSL_PROTOCOL = "TLSv1.2";
```
rjrudin, line 32: Just remove this and good to go.

rjrudin (approved):

PR #1744, opened 2025-04-07
Disable the TLSv1.3 tests when running in Java 8.

BillFarber:


rjrudin (dismissed):
Include a note - either in the commit description or a comment on the annotations - indicating why this is being done - i.e. because the version of Java 8 on Jenkins doesn't yet support TLSV1.3.

rjrudin (approved):

PR #1745, opened 2025-04-07
Enable TLSv1.3 with a different property.

BillFarber:
Trying this property to see if it works with Java 8.


marklogic-client-api/src/test/java/com/marklogic/client/test/ssl/OneWaySSLTest.java
```diff
@@ -130,6 +132,9 @@ void noSslContext() throws Exception {
 		);
 	}
 
+	// The TLS tests are failing on Java 8, because TLSv1.3 is enabled with our version of Java 8.
```
rjrudin, line 135: I think you mean "disabled with our version"?

rjrudin (dismissed):

PR #1746, opened 2025-04-25
Added tests for documenting how REST API patch works

rjrudin:
This was created many moons ago to help with designing Optic patch. The tests are all still relevant and are useful for documenting how the REST API patch operation works.


rjrudin:
Test failures are due to an issue in Optic in 12 nightly.

PR #1747, opened 2025-04-25
Updated test based on server fix for URIs with spaces

rjrudin:


BillFarber (approved):
Looks like that test is passing.

PR #1762, opened 2025-05-06
MLE-21383 Fixing "weak factory" Polaris issues

rjrudin:
Hoping to resolve "Improper Control of Document Type Definition" issues.


rjrudin:
This didn't make Polaris happy, and it got some failures on Jenkins that don't happen locally, so closing.

PR #1776, opened 2025-05-19
Enabling GraphQL tests

rjrudin:
These should be fixed soon in the server


rjrudin:
Eh, not doing this yet because this will cause all PR Jenkins test suites to fail as well. Will wait till the server has been fixed.

PR #1779, opened 2025-06-11
MLE-20906 Updated cosine and annTopK functions

rjrudin:
Based on changes to the server on 2025-06-06. `cosine-similarity` is now `cosine` and `annTopK` now accepts a map as its 5th argument.


BillFarber (approved):
Pending Jenkins success.

PR #1796, opened 2025-07-22
MLE-23004 - Found another Gradle deprecation.

BillFarber:


marklogic-client-api/build.gradle
```diff
@@ -62,6 +62,9 @@ dependencies {
 	testImplementation 'org.apache.commons:commons-lang3:3.17.0'
 	testImplementation 'com.opencsv:opencsv:5.10'
 	testImplementation 'org.skyscreamer:jsonassert:1.5.3'
+
+    testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.12.2")
```
rjrudin, line 66: Add a comment to explain why this is needed. There's also 1.13.4 - any reason not to use the latest version? (Maybe 1.13 requires Java 17 or higher?)

rjrudin (changes requested):

PR #1942, opened 2026-06-17
MLE-30240: Redact auth headers from OkHttp logs

jonmille:
## Summary

When OkHttp HTTP traffic logging is enabled via the
`com.marklogic.client.okhttp.httplogginginterceptor.level` system property,
`Authorization` and `x-auth-token` header values are now automatically
replaced with `[REDACTED]` before any log output is written — preventing
credentials from appearing in log files, stdout, or stderr.

## Changes

| File | Change |
|---|---|
| `OkHttpServices.java` | Added `redactSensitiveHeaders()` method; applied to all three log output paths; pre-compiled `Pattern` constant; security Javadoc added |
| `CONTRIBUTING.md` | New "HTTP Logging (Debugging Only)" section with security warning and property documentation |
| `RedactSensitiveHeadersTest.java` | New unit test (9 tests, no MarkLogic instance required) |

Jira Bug: https://progresssoftware.atlassian.net/browse/MLE-30240

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,10 +64,10 @@ The Java Client supports optional HTTP traffic logging via the system property
 
 > **Security warning:** `HEADERS` and `BODY` levels write HTTP headers and/or
 > request/response bodies to stdout, stderr, or your application logger.
-> `Authorization` header values are automatically redacted in the output, but
-> body content may include MarkLogic document data and must **never** be
-> enabled in production environments. Use these levels only during local
-> debugging sessions.
+> `Authorization` and `x-auth-token` header values are automatically redacted
+> in the output, but body content may include MarkLogic document data and must
+> **never** be enabled in production environments. Use these levels only during
+> local debugging sessions.
 
 Set the output target via `com.marklogic.client.okhttp.httplogginginterceptor.output`:
 `LOGGER` (routes to SLF4J at DEBUG level), `STDERR`, or leave unset for stdout.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,19 +55,3 @@ You can also undeploy the test application if you do not wish to keep it around 
 
 To test the `ml-development-tools` Gradle plugin, see the README.md file located at
 `ml-development-tools/src/test/example-project`.
-
-## HTTP Logging (Debugging Only)
-
-The Java Client supports optional HTTP traffic logging via the system property
-`com.marklogic.client.okhttp.httplogginginterceptor.level`. Accepted values are
-`BASIC`, `HEADERS`, `BODY`, and `NONE`.
-
-> **Security warning:** `HEADERS` and `BODY` levels write HTTP headers and/or
-> request/response bodies to stdout, stderr, or your application logger.
-> `Authorization` and `x-auth-token` header values are automatically redacted
-> in the output, but body content may include MarkLogic document data and must
-> **never** be enabled in production environments. Use these levels only during
-> local debugging sessions.
-
-Set the output target via `com.marklogic.client.okhttp.httplogginginterceptor.output`:
-`LOGGER` (routes to SLF4J at DEBUG level), `STDERR`, or leave unset for stdout.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
-To develop the Java Client and ultimately submit a pull request, you'll first need to be able to build the client and 
-run the tests locally. 
+To develop the Java Client and ultimately submit a pull request, you'll first need to be able to build the client and
+run the tests locally.
 
 Please see [this guide](.github/CONTRIBUTING.md) for information on creating and submitting a pull request.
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -53,5 +53,21 @@ You can also undeploy the test application if you do not wish to keep it around
...
+The Java Client supports optional HTTP traffic logging via the system property
+`com.marklogic.client.okhttp.httplogginginterceptor.level`. Accepted values are
+`BASIC`, `HEADERS`, `BODY`, and `NONE`.
+
+> **Security warning:** `HEADERS` and `BODY` levels write HTTP headers and/or
+> request/response bodies to stdout, stderr, or your application logger.
+> `Authorization` header values are automatically redacted in the output, but
+> body content may include MarkLogic document data and must **never** be
+> enabled in production environments. Use these levels only during local
+> debugging sessions.
```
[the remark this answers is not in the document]
jonmille: fixed

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -215,18 +239,24 @@ private OkHttpClient connect(ConnectionConfig config) {
...
 	 */
 	private void configureOkHttpLogging(OkHttpClient.Builder clientBuilder, Properties props) {
 		final boolean useLogger = "LOGGER".equalsIgnoreCase(props.getProperty(OKHTTP_LOGGINGINTERCEPTOR_OUTPUT));
 		final boolean useStdErr = "STDERR".equalsIgnoreCase(props.getProperty(OKHTTP_LOGGINGINTERCEPTOR_OUTPUT));
 		HttpLoggingInterceptor networkInterceptor = new HttpLoggingInterceptor(message -> {
-			if (useLogger == true) {
-				logger.debug(message);
-			} else if (useStdErr == true) {
-				System.err.println(message);
+			String redacted = redactSensitiveHeaders(message);
```
rjrudin: Something to consider while you're in here - this class is far too large. This custom interceptor would be easier to manage if it were in its own class, which then makes changes to it easier to understand. And it would also be easier to unit-test because you'd only be testing that class's redaction capabilities, you wouldn't have to mess with OkHttpServices at all.

CONTRIBUTING.md
```diff
@@ -53,5 +53,21 @@ You can also undeploy the test application if you do not wish to keep it around
 
     ./gradlew mlUndeploy -i -Pconfirm=true
 
-To test the `ml-development-tools` Gradle plugin, see the README.md file located at 
+To test the `ml-development-tools` Gradle plugin, see the README.md file located at
 `ml-development-tools/src/test/example-project`.
+
+## HTTP Logging (Debugging Only)
```
rjrudin: This isn't really appropriate in a CONTRIBUTING.md file, which is intended for developers working on this project. This is a user-facing feature that I don't think was ever properly documented in the official Java Client documentation. That's where this information needs to go. 

While this is useful to capture, I wouldn't put it here because it creates an expectation that an end user would need to look at this file, which they shouldn't ever have to care about. I think the better approach is to get this added to the official docs, probably here - https://docs.progress.com/bundle/marklogic-server-develop-with-java-12/page/topics/intro.html#id_98962

rjrudin (changes requested):

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
jonmille: Refactored the custom interceptor to RedactingHttpLogger.java.

CONTRIBUTING.md
jonmille: Removed the HTTP Logging section from the CONTRIBUTING.md file

marklogic-client-api/src/main/java/com/marklogic/client/impl/okhttp/RedactingHttpLogger.java
```diff
@@ -0,0 +1,59 @@
...
+	// Matches Authorization and x-auth-token header lines case-insensitively,
+	// capturing the full value through end-of-line. Uses .* (not \S+) so that
+	// multi-word values such as "Basic <token>", "Bearer <token>",
+	// "Negotiate <token>", and "Digest username=..." are fully captured.
+	// Java's . does not match \n, so adjacent lines are never affected.
+	static final Pattern SENSITIVE_HEADER_REDACTION_PATTERN =
+		Pattern.compile("(?i)(authorization|x-auth-token):.*");
+	static final String SENSITIVE_HEADER_REDACTION_REPLACEMENT = "$1: [REDACTED]";
+
+	private static final Logger logger = LoggerFactory.getLogger(RedactingHttpLogger.class);
```
rjrudin, line 33: Ah - this presents an easy way of addressing https://progresssoftware.atlassian.net/browse/MLE-29643 too . I do like the name "com.marklogic.client.okhttp.network" - I would use something like that instead of the class name so that the class name can be easily changed without impacting the logger.

rjrudin: That can be done in a separate PR - my point being that you might as well do that ticket for 8.2.0 as well since it will be so easy to do now.

rjrudin (approved):

marklogic-client-api/src/main/java/com/marklogic/client/impl/okhttp/RedactingHttpLogger.java
jonmille: I linked the 2 stories in Jira together with a relates to tag. We will revisit the story MLE-29643.

rjdew-progress (approved):

PR #1947, opened 2026-06-23
MLE-30242 remove .env from Git tracking and replace with .env.example

RitaChen609:


rjrudin (commented):
I don't think this is solving a problem yet. The only real issue I saw in MLE-30242 was that the name of the internal repository was identified. If that actually is a problem, then the Jenkinsfile needs to be updated as well. I am assuming the build is failing too because the Jenkinsfile was not updated and thus no value is found for `MARKLOGIC_IMAGE`.

I'm skeptical that that's really a problem though, as that internal repository name is in multiple GitHub repositories. I think some confirmation from a human is necessary here; MLE-30242 seems to be the result of an AI security scan. I would get that confirmation first before taking this step which makes life a little harder for both developers and for the CI process.

RitaChen609:
> I don't think this is solving a problem yet. The only real issue I saw in MLE-30242 was that the name of the internal repository was identified. If that actually is a problem, then the Jenkinsfile needs to be updated as well. I am assuming the build is failing too because the Jenkinsfile was not updated and thus no value is found for `MARKLOGIC_IMAGE`.
> 
> I'm skeptical that that's really a problem though, as that internal repository name is in multiple GitHub repositories. I think some confirmation from a human is necessary here; MLE-30242 seems to be the result of an AI security scan. I would get that confirmation first before taking this step which makes life a little harder for both developers and for the CI process.

You’re correct—this issue was identified as part of the AI security scan. I haven’t updated the Jenkinsfile yet, so it is still using a hardcoded internal repository for CI pipelines.

To address this, we should remove the hardcoded repository and make it configurable via an environment variable in Jenkins.

I’ve encountered a similar issue in [[MLE-30238] ](https://progresssoftware.atlassian.net/browse/MLE-30238) where we need to eliminate a hardcoded admin password. In that case, we’ll also need to introduce an environment variable (MARKLOGIC_ADMIN_PASSWORD) in Jenkins.

To ensure PR, nightly, and regression pipelines can continue to run automatically, these variables will likely need to be stored securely within the Jenkins credentials store.



[MLE-30238]: https://progresssoftware.atlassian.net/browse/MLE-30238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

RitaChen609:
I will remove the hardcoded repository and make it configurable via an environment variable in Jenkins in another PR along with the work for MLE-30238.

jonmille (approved):

RitaChen609:
We are not implementing this at present, as it would introduce additional complexity for both developers and the CI process.

PR #1963, opened 2026-08-03
MLE-31642 Update noSslContext Test

jonmille:
MarkLogic 12.1 is now returning a 403 error code, like we expect, instead of closing the connection and throwing a MarkLogicIOException for ML 12.1.

- Renamed noSslContext() -> noSslContextWithMarkLogic11OrLower()
- Added RequiresML12Dot0.java to only run tests against 12.0 versions
- Renamed noSslContextWithMarkLogic12() to noSslContextWithMarkLogic12Dot0()
- noSslContextWithMarkLogic12Dot0() ExtendWith updated to new RequiresML12Dot0.class to only run against 12.0 versions
- Added noSslContextWithMarkLogic12Dot1OrHigher() for new ML 12.1 behavior

Jira Ticket: https://progresssoftware.atlassian.net/browse/MLE-31642

### DIFF
--- a/marklogic-client-api/src/test/java/com/marklogic/client/test/junit5/RequiresML12Dot0.java
+++ b/marklogic-client-api/src/test/java/com/marklogic/client/test/junit5/RequiresML12Dot0.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2010-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+ */
+package com.marklogic.client.test.junit5;
+
+import com.marklogic.client.test.Common;
+import com.marklogic.client.test.MarkLogicVersion;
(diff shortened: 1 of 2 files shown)


rjrudin (changes requested):
This is definitely a good change by the server, but I think we need to keep a "RequiresML11OrLower" test that verifies that an IO exception is expected instead. While that's not desirable behavior, it's important to document that in one of our tests.

jonmille:
> This is definitely a good change by the server, but I think we need to keep a "RequiresML11OrLower" test that verifies that an IO exception is expected instead. While that's not desirable behavior, it's important to document that in one of our tests.

It IO exception was occurring for ML 12.0.x versions. The tests now cover the behaviors for 11 and lower, 12.0.x, and 12.1 and greater

rjrudin (approved):

rjdew-progress (approved):

PR #1935, opened 2026-05-01
MLE-28498 11.3.5 Test Fixes for Security Update

jonmille:
This PR primarily contains privilege, role, and permission test fixes as a result of the MLS 11.3.5 security fixes. Thes fixes were also applied to 12.0.2 and 12.1.0. 

Jira Story: https://progresssoftware.atlassian.net/browse/MLE-28498

PR #1966, opened 2026-08-11
MLE-31877 revert xdmp-login changes

jonmille:
Reverting the xdmp-login changes introduced as part of: https://github.com/marklogic/java-client-api/pull/1935

The `unprotected-uri` and `unprotected-collections` privileges are still needed so they have been added to the `java-test-delete-graph` and `java-test-evaluator` roles

The `xdmp:xslt-invoke`, `xdmp:xslt-eval`, and `xdmp:value` privileges are still needed so they have been added to the `test-test-writer` role

Jira Ticket: https://progresssoftware.atlassian.net/browse/MLE-31877
Related issue why `xdmp-login` is being removed: https://progresssoftware.atlassian.net/browse/MLE-30664


test-app/src/main/ml-config/security/roles/test-rest-writer.json
```diff
@@ -18,13 +18,23 @@
 			"kind": "execute"
 		},
 		{
-			"privilege-name": "xdmp-login",
-			"action": "http://marklogic.com/xdmp/privileges/xdmp-login",
+			"privilege-name": "xdmp:invoke",
```
rjrudin, line 21: This is fine for this PR, but I think it'd be good to next try a PR where this is removed from this role and see what fails. And then create an additional role - e.g. "test-invoke" - that only grants this privilege and give it a user that is only used for tests where we know this privilege is required. It should be granted for a "rest writer" who typically does not need invoke (or eval) privileges.

test-app/src/main/ml-config/security/roles/test-rest-writer.json
```diff
@@ -18,13 +18,23 @@
...
+			"action": "http://marklogic.com/xdmp/privileges/xdmp-invoke",
 			"kind": "execute"
 		},
 		{
-			"privilege-name": "xdmp:invoke",
-			"action": "http://marklogic.com/xdmp/privileges/xdmp-invoke",
+			"privilege-name": "xdmp:xslt-invoke",
```
rjrudin, line 26: Same for these privileges - I would do a follow up PR where I remove these and see what breaks.

rjrudin (approved):

test-app/src/main/ml-config/security/roles/test-rest-writer.json
jonmille: With the removal of xdmp:xslt-invoke, xdmp:xslt-eval, and xdmp:value the 3 test failures from the first Jenkins run are shown: https://ml-clt-jenkins.progress.com/blue/organizations/jenkins/devexp%2FJava-Client%2Fjava-client-api/detail/PR-1966/1/tests

xsltTransformWithoutParam() – com.marklogic.client.test.rows.TransformDocTest
- `com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to apply resource at rows/update. Server Message: You do not have permission to this method and URL.`

xsltTransformWithParam() – com.marklogic.client.test.rows.TransformDocTest
- `com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to apply resource at rows/update. Server Message: You do not have permission to this method and URL.`

testUsingFromParamAndSchematron() – com.marklogic.client.test.rows.ValidateDocTest
- `com.marklogic.client.ForbiddenUserException: Local message: User is not allowed to apply resource at rows/update. Server Message: You do not have permission to this method and URL.`

jonmille: - `TransformDocTest#xsltTransformWithoutParam` — XSLT stylesheet invoked by URI; needs `xdmp:xslt-invoke`
- `TransformDocTest#xsltTransformWithParam` — XSLT stylesheet invoked by URI; needs `xdmp:xslt-invoke`
- `ValidateDocTest#testUsingFromParamAndSchematron` — schematron compiled to XSLT (`xdmp:xslt-invoke` + `xdmp:xslt-eval`) then written via `fromParam` (`xdmp:value`)

rjrudin: That's great - I wouldn't address those in this PR, but I would do a follow up where I make a new test user that has those additional privileges. That way, the other tests run as a rest-writer user that doesn't have those privileges. That provides a better regression suite that can catch a bug where a server change ends up requiring a new privilege.

RitaChen609 (approved):

Issue #1938, opened 2026-05-20
Version number missing on DocumentDescriptor


ralfhergert:
We are relying on the MarkLogic server OptimisticLocking feature. It works via the version number in the DocumentDescriptor: when a write attempt is made via `XmlDocumentManager.write(DocumentDescriptor, writeHandle)` a `ContentWrongVersionException` is throw when the server discovered that the document did change since the last read attempt.

We are currently looking for a way to read the content of a document and its DocumentDescriptor in a single server request and discovered the following inconsitency. While
```
XmlDocumentManager.exists(String uri)
```
returns a DocumentDescriptor with the proper version number being set
```
XmlDocumentManager.read(String uris...)
```
returns a DocumentPage with DocumentRecords with DocumentDescriptors in which the version number is always -1.

Request is: Could we please also get the correct document version number when XmlDocumentManager.read(String uris...) is being used?

(We are using the marklogic-client in version 8.1.0 and the MarkLogic server in version 11.3.4)

rjdew-progress:
Tracked internally with JIRA issue https://progresssoftware.atlassian.net/browse/MLE-29883

ralfhergert:
Hi @rjdew-progress, any news on this issue?

rjdew-progress:
Hi @ralfhergert. Apologies for the delay in response. The fix for this issue will be in our 8.2.0 release currently planned for mid to late September.

rjdew-progress:
Closing as the issue has been resolved in the 8.2.0 release

PR #1968, opened 2026-08-17
MLE-29883 (GH #1938) Include document version in bulk reads

rjdew-progress:
# Impact

Users who rely on MarkLogic optimistic locking cannot use DocumentManager.read(String uris...) to get a usable document version number. This can prevent correct version checks before writes and affect concurrent update handling.

# Expected behavior

DocumentManager.read(String uris...) should return DocumentRecords with DocumentDescriptors that include the correct document version number, like DocumentManager.exists(String uri) does.

# Actual behavior

DocumentManager.read(String uris...) returns a DocumentPage with DocumentRecords whose DocumentDescriptors always have the version number set to -1.

# Steps to reproduce

Use DocumentManager.exists(String uri).

Observe that it returns a DocumentDescriptor with the proper version number set.

Use DocumentManager.read(String uris...).

Observe that it returns a DocumentPage with DocumentRecords whose DocumentDescriptors have the version number set to -1.

GitHub issue #1938

### DIFF
--- a/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
+++ b/marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
@@ -19,6 +19,7 @@ import com.marklogic.client.document.DocumentManager.Metadata;
 import com.marklogic.client.eval.EvalResult;
 import com.marklogic.client.eval.EvalResultIterator;
 import com.marklogic.client.extra.okhttpclient.OkHttpClientConfigurator;
+import com.marklogic.client.impl.okhttp.HeaderUtil;
 import com.marklogic.client.impl.okhttp.HttpUrlBuilder;
 import com.marklogic.client.impl.okhttp.OkHttpUtil;
 import com.marklogic.client.impl.okhttp.PartIterator;
@@ -36,9 +37,7 @@ import com.marklogic.client.util.RequestParameters;
 import jakarta.mail.BodyPart;
 import jakarta.mail.Header;
 import jakarta.mail.MessagingException;
-import jakarta.mail.internet.ContentDisposition;
 import jakarta.mail.internet.MimeMultipart;
-import jakarta.mail.internet.ParseException;
 import jakarta.mail.util.ByteArrayDataSource;
 import jakarta.xml.bind.DatatypeConverter;
 import okhttp3.*;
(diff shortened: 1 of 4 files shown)


jonmille (dismissed):

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -1713,6 +1713,17 @@ static private Format getHeaderFormat(BodyPart part) {
 		return null;
 	}
 
+	// Bulk multi-document reads carry the version as a "versionId" param on Content-Disposition, not as an ETag header.
+	static private long getHeaderVersion(BodyPart part) {
```
rjrudin: One thought on where to put this - I tried to avoid adding anything to `OkHttpServices` because it's so large already. The trick is where to put it instead, as there's not a good pattern for that. 

Nonetheless, I would still lean towards a separate class in the `impl.okhttp` subpackage. It could be as simple as `ContentDispositionUtil` with a `public static String getVersion`. One small bonus would be that it provides a nice home for declaring the version as a private static, e.g. `private static final VERSION_PATTERN = "..."`. 

Could also add the ETag fallback, which seems useful since that was the existing behavior - I can't say for sure, but maybe an older version of MarkLogic returned it as an ETag?

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
```diff
@@ -4688,7 +4702,7 @@ public DocumentDescriptor getDescriptor() {
 			updateFormat(descriptor, getFormat());
 			updateMimetype(descriptor, getMimetype());
 			updateLength(descriptor, getLength());
-			updateVersion(descriptor, content.getHeader(HEADER_ETAG));
+			updateVersion(descriptor, content.getVersion());
```
rjrudin, line 4512: Yeah, seeing this existing code here makes me think that Copilot's suggestion is a good one about using the ETag as a fallback.

marklogic-client-api/src/main/java/com/marklogic/client/impl/OkHttpServices.java
rjdew-progress: Moved to OkHttpUtil and now fallback to the ETAG header.

marklogic-client-api/src/main/java/com/marklogic/client/impl/okhttp/OkHttpUtil.java
```diff
@@ -187,6 +207,209 @@ private static void initializeSslContext(OkHttpClient.Builder clientBuilder, SSL
...
+			format = contentDisposition.replaceFirst("^.*" + formatRegex + ".*$", "$1");
+			return Format.valueOf(format.toUpperCase());
+		} else if (contentType != null && contentType.length() > 0) {
+			return Format.getFromMimetype(contentType);
+		}
+		return null;
+	}
+
+	// Bulk multi-document reads usually carry the version as a "versionId" param on Content-Disposition.
+	public static long getHeaderVersion(BodyPart part) {
```
rjrudin: Sorry, this is the only part that I meant moving. I actually would keep the rest in OkHttpServices for now. It could all use a huge refactor to simplify it. My point before was only that I try not to add to it if possible. So I would keep all the other methods in there for now and just put the new code somewhere else. 

I actually wouldn't put it in this class, which at least has a small scope for now of creating a client builder object. Maybe a HeaderUtil instead - something with a scope of processing headers from MarkLogic?

rjdew-progress: Moved the functions to a separate HeaderUtil class. I still moved other header related functions since they were private, so there isn't a risk of a user being dependent on the them and it'll clean up the OkHttpServices a bit.

RitaChen609 (dismissed):

jonmille (approved):

Issue #1969, opened 2026-08-19
MarkLogicServerException.getServerStatusCode returns NullPointerException


fromanu:
MarkLogicServerException.getServerStatusCode returns NullPointerException in case the failedRequest is null:
`  public int getServerStatusCode() {
    return (failedRequest == null) ? null : failedRequest.getStatusCode();
  }`
because the return type of the method is a primitive int.
Stack trace:
`java.lang.NullPointerException: Cannot invoke \"java.lang.Integer.intValue()\"\n\tat com.marklogic.client.MarkLogicServerException.getServerStatusCode(MarkLogicServerException.java:54)`

(We are currently using the marklogic-java-client in version 8.1.0 and the Markogic Server in 11.3.6)

SoumenduGhosh10:
This looks like an issue in the product. We have created JIRA item https://progresssoftware.atlassian.net/browse/MLE-32022 to track and fix the issue

fromanu:
That's great. Thanks a lot for the quick reaction and for tackling it.

rjdew-progress:
This has been fixed in the 8.2.0 release

PR #1972, opened 2026-08-21
MLE-32022 Avoid null pointer exception when FailedRequest is null

rjdew-progress:
Addresses GitHub issue [#1969](https://github.com/marklogic/java-client-api/issues/1969)


rjrudin (approved):
An interesting design choice as to what to return! I could see an argument for returning 500 - i.e. "Well we know the server complained about something, but we have no idea why for internal reasons, so here you go, have a 500" - but -1 seems reasonable as well as a way of communicating that we don't know what happened.

ngodugu-marklogic (approved):

RitaChen609 (approved):

